PR #2, opened 2017-02-21
Multiple in-order host tasks with data guarantees

ProGTX:
It should be possible to have multiple host tasks in the same command group and their execution should be in-order.


Ruyk:
This makes sense, and follows the rules of the separate section of multiple kernels per command group. Thanks!

PR #3, opened 2017-03-02
Disallow mixing host tasks with kernel invocations

ProGTX:
Since there can currently only be one kernel per command group, there are certain expectations that need to be met and it would be difficult to meet them if host tasks were allowed alongside kernel invocations in the same command group. So until the limit of one kernel per group is lifted, this should should be disallowed, i.e. even though multiple host tasks per command group are allowed, also using a kernel invocation in the same command group should not be.


Ruyk:
Although it makes sense to have this limit, I believe we should have a more user-friendly approach. A separate handler for host and device API calls would simplify things, as you'll only be able to use one or the other

ProGTX:
I've added the definition of `host_handler` similar to how the `handler` is defined in the spec and that it can be used in functions that expect a normal handler.

Ruyk:
Looks fine to me

PR #8, opened 2017-06-05
P0567r1

AerialMantis:
* Update proposal based on feedback.
* See change list in document for full list of changes.


managed-pointer/cpp-20/D0567R1.md
```diff
@@ -5,222 +5,384 @@
...
+#### 3.3.3. Memory accessibiity
 
-As described previously there are three ways in which a synchronisation operation can be triggered, and these are separated into explicit and implicit.
+During the lieftime of a `managed_ptr` it's managed memory allocation can exist in the memory region(s) associated with any number of *execution contexts*, but may only be accessible in one of these memory regions at any given time. The *execution context* who's memory region is currently accessible is said to be the accessible *execution context*.
 
-The first is by calling one of the member functions or global customisation point functions described above. This will trigger a synchronisation operation to the memory region of the **execution context** associated with the **executor**. This method is useful as it allows finer control over the synchronisation operations by chaining continuations. This is implicit.
+> This limited may be relaxed in a future version of the paper when a `managed_ptr` can be concurrently accessed in more than one memory region. In this case there could be more than one accessible memory region. This is further described in section 5.1.
```
Ruyk: Note that this will likely require hardware support for atomic operations of some sort, which is effectively a hardware capability

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -5,222 +5,384 @@
...
 
-The second is by calling the ``get()`` member function on the ``managed_ptr`` itself as this will implicitly trigger a call to ``get()`` on the current accessible **execution context** if the host CPU is not currently accessible. This method is useful when you want to synchronise data back to the host CPU in order to use the resulting data in regular sequential code. This is explicit.
+The `managed_ptr` must maintain a reference to the current accessible *execution context*.
```
Ruyk: a reference or a copy? Can the execution context instance be deleted or has to be alive? There is a fundamental difference in the sense that if an execution context destructor has some expected behaviour (e.g synchronization), which will be triggered then by a chain of destructors from the managed_ptr if a copy is stored, but not if there is a reference.

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -5,222 +5,384 @@
...
 
-The third is by passing a ``managed_ptr`` to an executor control structure such as ``async()``, as this will implicitly trigger a call to the ``put()`` member function on the **executor**. If the host CPU is not accessible at this point then this will implicitly trigger a call to``get()`` on the ``managed_ptr`` which will subsequently implicitly trigger a call to ``get()`` on the currently accessible **execution context**. This method is useful as it allows you to simply pass a  ``managed_ptr`` directly to a control structure without requiring an explicit call to ``put()``. This is implicit.
+The associated *execution context* and *executor* of a `managed_ptr` can also be accessible, the `managed_ptr` has a member function for querying wether the associated *execution context* and *executor* is accessible; `is_accessible()`.
 
-## Examples
+The data at the memory region of the `managed_ptr`s associated *execution context* and *executor* can be read and modified (if the type `T` is a non-const type) via the de-reference and pointer operators.
```
Ruyk: does that mean it is possible to get the address of that pointer? If so, is pointer arithmetic valid?

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -5,222 +5,384 @@
...
 
-The first example shows how this would look using the explicit interface for synchronisation operations (figure 3) and the second shows how this would look using the implicit interface for synchronisation operations (Figure 4).
+If a user wishes to explicitly allocate or deallocate memory for a `managed_ptr` instance on a particular execution context, this can be done by calling the allocation operations `allocate`, `then_allocate`, `deallocate` and `then_deallocate`. Each of which is an asycnhronous operation which performs an allocation or dallocation on a particular execution context and returns a future to the complletion of the operation.
```
Ruyk: wishes -> wants
asycnhronous -> asynchronous
Each of which -> "Each function"
complletion -> "completion"

Ruyk: What guarantees of explicit allocation are provided by the interface? Can an implementation decide to do a no-op on these functions if it is not required?

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -5,222 +5,384 @@
...
+
+If any memory allocation or deallocation operations fail then an exception is thrown and stored in the future object that is returned by the operations which triggered the allocation operation.
+
+#### 3.3.5. Synchronisation
+
+For an **execution context** to be the accessible **execution context**, if it is not already, a synchronisation operation is required. A synchronisation operation is an implementation defined asynchronous st of commands which moves the data from the currently accessible memory region to another memory region which returns a future object that can be used to wait for the operation to complete. From the point at which a synchronisation operation is triggered the currently accessible memory region is no longer accessible. Once the synchronisation point is complete (i.e. the future returned from the synchronisation operation is complete) the memory region the data is being synchronised to is now the accessible memory region.
```
Ruyk: "asynchronous st of commands" -> asynchronous set of commands

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -5,222 +5,384 @@
...
+
+Synchronisation operations are course grained synchronisation in that they synchronise the entire managed memory allocation of a `managed_ptr`.
```
Ruyk: course -> coarse

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -5,222 +5,384 @@
...
+
+> This limited may be relaxed in a future version of the paper when a `managed_ptr` can be concurrently accessed in more than one memory region. In this case there sychronisation could be performed at smaller granularities using mapping operations or atomic operations. This is further described in section 5.1.
+
+There are three sycnhronisation operations; `get`, `put` and `sync`, each taking one or more `managed_ptr` parameters. These sycnhronisation operations perform the following sycnhronisation for each `managed_ptr` parameter:
```
Ruyk: sycnhronisation ->  synchronization 
sycnhronisation ->  synchronization

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -5,222 +5,384 @@
...
+
+* The `get` operation synchronises the data at the memory region at a particular *execution context* to the memory region at the *execution context* associated with the `managed_ptr`.
+* The `put` operation synchronises the data at the memory region at the *execution context* associated with the `managed_ptr` to the memory region at a particular *execution context*.
+* The `sync` operation synchronises the data at the memory region at one particular *execution context* to the memory region at another particular *execution context*.
```
Ruyk: If the managed_ptr knows where it is being used, why do you need explicit two-way operations? Seems you only need a "sync" to perform a manual update

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -5,222 +5,384 @@
...
+
+> In the previous revision of this paper it was also possible to implicitly trigger sycnhornisation operations by simplly passing a `managed_ptr` to a control structure such as `async`. However this has been removed due to a limitation that may effect implementations. This is described in section 5.2.
```
Ruyk: simplly -> simply
may effect -> may affect

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -5,222 +5,384 @@
...
+
+There are a further three sycnhronisation operations; `then_get`, `then_put` and `then_sync`, which are equivelants of `get`, `put` and `sync` but take a predecate future parameter that the operation must wait on before executing. These are used to provide a continuations interface for sycnhronisation operations. Equivelant operations to these are also required as member functions on the future type, this is described further in section 3.4.
```
Ruyk: take a predecate -> take a predicate? 
synchronisation -> synchronization
Equivelant -> equivalent
"type, this is" -> type. This is"

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -5,222 +5,384 @@
...
+
+If any synchronisation operations fail then an exception is thrown and stored in the future object that is returned by the operations which triggered the synchronisation operation.
```
Ruyk: "fail" -> "fails,"

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -229,94 +391,105 @@ auto gpuExecutor = gpuContext.executor();
 auto gpuAllocator = gpuContext.allocator();
 
 /* Construct a managed_ptr ptraA allocated on the host CPU */
-std::experimental::execution::managed_ptr<float> ptrA(1024);
+execution::managed_ptr<my_data> ptrA();
```
Ruyk: The empty parenthesis are not needed.
Does this calls the constructor? 
If so, is the constructor called on each allocation on different memory spaces or only once and then replicated?
Depending on this, the managed_ptr may be restricted to trivial constructable types or not, but it is not specified

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -229,94 +391,105 @@ auto gpuExecutor = gpuContext.executor();
...
 populate(ptrA);
 
 /* Construct a series of compute and data operations */
 auto fut =
-  std::experimental::execution::put(gpuExecutor, ptrA)
-      .then_put(gpuExecutor, ptrB)
-          .then_async(gpuExecutor, [=](auto _ptrA, auto _ptrB) { /* ... */ }, ptrA, ptrB)
-            .then_get(gpuExecutor, ptrA, ptrB);
+  execution::put(gpuExecutor, ptrA)
+    .then_async_execute(gpuExecutor, [=]() { /* ... */ })
```
Ruyk: what will happen if a different executor is used on this then_async_execute?

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -229,94 +391,105 @@ auto gpuExecutor = gpuContext.executor();
...
+      .then_get(ptrB);
```
Ruyk: Does the get imply a copy to the memory handled by the managed_ptr on the *host* according to the previous definitions?

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -229,94 +391,105 @@ auto gpuExecutor = gpuContext.executor();
...
 fut.wait();
 
 /* Print the result */
 print(ptrB);
 ```
+*Figure 2: Example of synchronisation operations*
+### 3.4. Requirements on the Future Type
 
-*Figure 3: Example of explicit interface*
+The executor future type `executor_future_t` that is either the result of or calls a sycnhronisation continuation must provide the member functions `put`, `get`, `sync`, `allocate` and `deallocate` as described in the execution synopsis (Figure 1).
```
Ruyk: synchronisation continuation -> synchronization continuation

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -229,94 +391,105 @@ auto gpuExecutor = gpuContext.executor();
...
 
-```cpp
-/* Construct a context and executor for executing work on the GPU */
-gpu_execution_context gpuContext;
-auto gpuExecutor = gpuContext.executor();
+### 3.5. Optimising Sychronisation Operations using Synchronisation Channels
 
-/* Retrieve gpu allocator */
-auto gpuAllocator = gpuContext.allocator();
+Synchronisation channels are objects which represent a connection between two execution contexts, for example between two remote devices or between an I/O device and a remote device. The default sycnhronisation channel which supports sycnhronisation between all execution contexts is that which sycnhronises via the host execution context. However implementors can provide their own synchronisation channels which can be used by the user in synchronisation operations to optimise them.
```
Ruyk: synchronises -> synchronizes

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -229,94 +391,105 @@ auto gpuExecutor = gpuContext.executor();
...
-## Future Work
+## 5. Future Work
 
 There are many other considerations to make when looking at a model for data movement for heterogeneous and distributed systems, however, this paper aims to establish a foundation which can be extended to include other paradigms in the future.
 
-### More Complex Data Movement Policies
+### 5.1. Shared Accessibility
 
-We may wish to introduce more control over the way in which data is moved between devices at a high level and the consistency between execution contexts. This could be done by having static type traits associated with executor classes which describe the data movement and consistency properties of the execution context associated with the executor.
+Systems which support a unified address space (i.e. sharing a single memory region across different *execution contexts*) could allow for a `managed_ptr` instance to share accessibility between multiple *execution contexts*. This would allow a `managed_ptr` to be concurrently accessible by multiple *execution contexts* via the use of atomics providing those *execution contexts* share a memory region.
```
Ruyk: Do you even need the managed_ptr if your system supports concurrent memory access? In this systems, the managed_ptr  can be just a wrapper to a normal pointer

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -229,94 +391,105 @@ auto gpuExecutor = gpuContext.executor();
...
+* The memory mode `memory_mode::discrete` can only have a single accesible execution context so cannot be accessed concurrently and has discrete emory regions so cannot support passing pointer based structures.
+* The memory mode `memory_mode::unified` can only have a single accesible execution context so cannot be accessed concurrently but has a unified memory region so can support passing pointer based structures.
+* The memory mode `memory_mode::shared` can have multiple accesible execution contexts so can be accessed concurrently via the use of atomics and has a unified memory region so can support passing pointer based structures.
 
-### Synchronisation Operations From Multiple Host CPU Devices
+This proposal is not yet in the main paper and so is considered future work as the idea has not yet been fully fleshed out and there are other factors that have to be considered such as the execution context - memory region topology and atomic operations.
```
Ruyk: "This proposal is not yet in the main paper" ?

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -229,94 +391,105 @@ auto gpuExecutor = gpuContext.executor();
...
+In a previous revision of this paper it was possible to trigger implicit sycnhronisation operations by simply passing a `managed_ptr` to a control structure such as `async`. However this has been removed due to limitations that may effect an implementation. As the current executors interface does not support variadic execution function parameters, implicit an implementation would be required to trigger sycnrhonisation operations on all control structures for `managed_ptr` captures. Firstly this may incur unwanted overhead and secondly this could potentially require static reflection for an optimal implementation. For these reasons this feature has been removed and instead kept as a potential future work that could be incorporated if proven to be implementable without creating overhead.
```
Ruyk: syncrhonisation -> synchronization

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -229,94 +391,105 @@ auto gpuExecutor = gpuContext.executor();
...
 
-### Additional Containers
+### 5.3. Additional Containers
 
-We may wish to extend this principle of a managed pointer to other containers that would be useful to share across heterogeneous and distributed systems such as vectors or arrays. This could be done by having containers such as ``managed_vector`` or ``managed_array`` that would have similar requirements to the standard containers of the same names in terms of storage and access yet would be extended to support access from remote devices as with the ``managed_ptr``.
+We may wish to extend this principle of a managed pointer to other containers that would be useful to share across heterogeneous and distributed systems such as vectors or arrays. This could be done by having containers such as `managed_vector` or `managed_array` that would have similar requirements to the standard containers of the same names in terms of storage and access yet would be extended to support access from remote devices as with the `managed_ptr`.
```
Ruyk: Seems that you should be able to build containers on top of the managed_ptr , so no need for complicating the interface more

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -229,94 +391,105 @@ auto gpuExecutor = gpuContext.executor();
...
 
-## Data Movement Customisation Points
+### 5.4. Hierarchical Memory Structures
```
Ruyk: Structures -> Architectures

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -5,222 +5,384 @@
...
+* Expand on the section about work flow models.
 
-The approach proposed in this paper does not include all of the use cases that a complete solution would cover.
-* This approach makes the assumption that there is only a single host CPU device which is capable of performing synchronisation with **execution contexts**.
-* This approach does not include an optimized manner of moving data between **execution contexts**.
-* This approach does not include support for systems which allows for multiple devices to access the same memory concurrently.
+## Feedback
 
-## Introduction
+Much valuable feedback was given dsince the last revision of this paper, this section will identify the issues that were raised and points to their respective solutions.
```
AerialMantis: Fixed.

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -5,222 +5,384 @@
...
 
-In non-heterogeneous or non-distributed systems, there is typically a single device: a host CPU with a single memory region or an area of addressable memory. In contrast, heterogeneous and distributed systems have multiple devices (including the host CPU) with their own discrete memory regions.
+It was suggested that the `managed_ptr` should be capable of providing read-only access, where write access is restricted, this can be very useful as it allows an implementation to make use of read-only partitions of memory which are often more efficient to access. This is now supported in this paper and is described in section 3.3.
 
-A device in this respect can be any architecture that exists within a system that is C\+\+ programmable; this can include CPUs, GPUs, APUs, FPGAs, DSPs, NUMA nodes, I/O devices and other forms of accelerators.
+### Allocation of managed memory
+
+There were many questions regarding how the how and when memory is allocated for a `managed_ptr` instance and what happens on an allocation failure. Section 3.3.4 has been added which describes these behaviours.
+
+### Optimising sycnhornisation between execution contexts
```
AerialMantis: Fixed.

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -5,222 +5,384 @@
...
+
+It was suggested that it be possible for an implementation to optimise synchronisation of a `managed_ptr` between particular execution contexts rather than having to synchronise via a host execution context. A solution that was being discussed but didn't make it into the first revision of the paper was the concept of channels. This is now supported in this paper and is described in section 3.5.
```
AerialMantis: Fixed.

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -5,222 +5,384 @@
...
+
+## 1. Introduction
+
+### 1.1. Summary
+
+This paper proposes an extension to the unified interface for execution proposal [1] to facilitate the management and sycnchronisation of a memory allocation across multiple memory region(s). This addition is in the form of the class template `managed_ptr`; similar to the `std::shared_ptr` but with the addition that it can share its memory allocation across the memory region of it's host execution context where it is constructed and the memory region(s) of one or more other host or remote execution contexts.
+
+The aim of this paper is to begin an exploratory work into designing a unified interface for data movement. There are many different architectures and data flow models to consider when designing such an interface so it is expected that this paper will serve only as a beginning.
```
AerialMantis: Done.

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -5,222 +5,384 @@
...
 
-Alternative names that were considered for `managed_ptr` were `temporal` as it described a container which gave temporal access to a managed memory allocation, `managed_container` as the original design was based on the `std::vector` container and `managed_array` as the managed memory allocation is statically sized.
+### 2.1. Must support different memory architectures
 
-Alternative names for the `put()` and `get()` interface were `acquire()` and `release()` as you were effectively acquiring and releasing the managed memory allocation and `send()` and `receive()` as you are effectively sending and receiving back the managed memory allocation.
+In non-heterogeneous or non-distributed systems, there is typically a single device: a host CPU with a single memory region or an area of addressable memory. In contrast, heterogeneous and distributed systems have multiple devices (including the host CPU) often with their own discrete memory regions. A device in this respect can be any architecture that exists within a system that is C\+\+ programmable; this can include CPUs, GPGPUs, APUs, FPGAs, DSPs, NUMA nodes, I/O devices and other forms of accelerators.
```
AerialMantis: Fixed.

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -5,222 +5,384 @@
...
+
+These memory regions can be categorised into two groups. Firstly memory architectures which have a single unfied address space, that is a single area of addressable memory that is accessible across multiple devices, this can optionally be supprted in hardware by devices having the same physical memory or in software by having a cache coherancy layer accessing a shared virtual address space. Secondly memory architectures which have multiple address spaces, therefore having separate discrete areas of addresable memory, often in different physical memory.
+
+A unified interface needs to be capable of supporting both of these kinds of memory architectures in such a way that any requirements for cache coherancy is abstracted by the interface and associated coherency model.
```
AerialMantis: So this is something that is being considered for systems with shared memory regions, however this is not yet covered by the proposal. This is currently described in the future work section in "5.1. Shared Accessibility".

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -5,222 +5,384 @@
...
 
-The **execution context** must encapsulate both the execution agents and a global memory region that is accessible from those execution agents. It is required that the memory region of the **execution context** be cache coherent across the execution agents created from the same invocation of an execution function. It is not required that the memory region of the **execution context** be cache coherent across execution agents created from different execution functions or on different **execution contexts**.
+Another data flow model is one that uses a runtime to create a graph of tasks, which can be optimised.
 
-*Note: some systems are capable of providing cache coherency across execution functions or **execution contexts** through the availability of shared virtual memory addressing or shared physical memory. This feature is discussed in more detail in the future work section.*
+Another data flow model is one which simply submits work to be executed and waits for it all to complete, typically called fire and forget execution.
 
-### Managed Pointer Class Template
+### 2.4. Must integrate seemlessly with exexcutors
```
AerialMantis: Fixed.

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -5,222 +5,384 @@
...
 
-The proposed addition to the standard library is the ``managed_ptr`` class template (Figure 1). The ``managed_ptr`` is a smart pointer which has ownership of a contiguous managed allocation of memory that is shared between the host CPU and one or more **execution contexts**. It is important to note here that an **execution context** may be on the host CPU accessing the host CPU memory region, however, the interface remains the same.
+As data movement is tightly coupled with executtion, it is nessesary for a unified interface for data movement to be integrated with the unified interface for execution. This means such a unified interface must integrate seemlessly with the current design philosophy of executors, in that it must be compatible for the range of control structures which the executors abstraction supports and must support te range of platforms which executors can target.
```
AerialMantis: Fixed.

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -5,222 +5,384 @@
...
+
+It was suggested that it be possible for an implementation to optimise synchronisation of a `managed_ptr` between particular execution contexts rather than having to synchronise via a host execution context. A solution that was being discussed but didn't make it into the first revision of the paper was the concept of channels. This is now supported in this paper and is described in section 3.5.
+
+### Discuss memory access in terms of side effects
+
+It was suggested that this proposal should describe memory access in terms of visible sided effects rather than caching. This has been introduced as part of the requirement for supporting different memory architectures and the paper has been updated to reflect this.
+
+### Host based and remote execution contexts
+
+It was asked what would happen if `malloc()` were to be called on a GPGPU, and the answer is this cannot be supported due to hardware limitations on some devices such as GPUs. This question of *execution context* capabilities is described in section 3.3.5.
```
AerialMantis: That's a good point, I meant to say for the general case, i've updated this.

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -5,222 +5,384 @@
...
+
+The aim of this paper is to begin an exploratory work into designing a unified interface for data movement. There are many different architectures and data flow models to consider when designing such an interface so it is expected that this paper will serve only as a beginning.
+
+### 1.2. Motivation
+
+In non-heterogeneous or non-distributed systems, there is typically a single device: a CPU with a single memory region or an area of addressable memory. In contrast, heterogeneous and distributed systems have multiple devices with shared or discrete memory regions.
+
+A device in this respect can be any architecture that exists within a system that is C\+\+ programmable; this can include CPUs, GPGPUs, APUs, FPGAs, DSPs, NUMA nodes, I/O devices and other forms of accelerators.
```
AerialMantis: Done.

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -5,222 +5,384 @@
...
+In non-heterogeneous or non-distributed systems, there is typically a single device: a host CPU with a single memory region or an area of addressable memory. In contrast, heterogeneous and distributed systems have multiple devices (including the host CPU) often with their own discrete memory regions. A device in this respect can be any architecture that exists within a system that is C\+\+ programmable; this can include CPUs, GPGPUs, APUs, FPGAs, DSPs, NUMA nodes, I/O devices and other forms of accelerators.
+
+These memory regions can be categorised into two groups. Firstly memory architectures which have a single unfied address space, that is a single area of addressable memory that is accessible across multiple devices, this can optionally be supprted in hardware by devices having the same physical memory or in software by having a cache coherancy layer accessing a shared virtual address space. Secondly memory architectures which have multiple address spaces, therefore having separate discrete areas of addresable memory, often in different physical memory.
```
AerialMantis: Done.

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -5,222 +5,384 @@
...
 
-At any one time the managed memory allocation can exist in the memory regions of the host CPU and any number of **execution contexts**. However, the managed memory allocation may only be accessible in one of these memory regions at any given time. This memory region is said to be the accessible memory region. If the accessible memory region is of the host CPU, the host CPU is said to be accessible and if the accessible memory region is of an **execution context** that **execution context** is said to be the accessible **execution context**.
+At the same time a unified interface for data movement must exist at a high level of abstraction and must avoid introducing too many requirements on executor implementations, in order to avoid introducing complex requirements for heterogeneous or distributed memory architectures into C\+\+.
```
AerialMantis: So the intention of the `managed_ptr`, in the current proposal (not taking into consideration systems with a unified address space) is that once synchronized, it will be accessible to all execution agents of the execution context concurrently within a single execution. Currently there is no guarantees made for consistency between execution agents concurrently accessing the same address of the `managed_ptr`, i.e. the user is responsible for avoiding race conditions. The reason for this is that the exact behavior will vary from one device to another and cannot yet be fully described by the C++ memory model, so this is something that I intended to discuss in greater detail as the capabilities of devices is fleshed out in the executors proposal. However we should mention this in future work as it is very important to take into consideration.

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -5,222 +5,384 @@
...
+
+Different memory architectures, both unified and discrete; may support granularity at different levels. This can either be at the course grained level; where synchronisation between memory regions is performed at the execution function level, or this can be at the fine grained level; where synchronisation between memory regions is performed at the instruction level, allowing access to memory from different devices concurrently, using atomics to specify read and write order.
+
+A unified interface needs to be capable of supporting both these levels of synchronisation.
 
-## Proposed Additions
+### 2.3. Must support different data flow models
 
-### Requirements on Execution Context
+There are also many different data flow models that can be used to describe data and data dependenies in relation to functions of execution. Though it would be impossible to cover all possible models, it's important to make a unified inteface as flexible and utilisable as possible, therfore facilitating a wide range of data flow models.
```
AerialMantis: I have expanded on this section to describe what is meant by data-flow model and the kinds of data-flow models that the proposal should support.

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -5,222 +5,384 @@
...
 
-For the purposes of this paper, it is necessary to introduce requirements on the **execution context**; the object associated with an **executor** which encapsulates the underlying execution resource on which functions are executed on.
+One such data flow model that C\+\+ is moving towards is that of futures and continuations; the idiom that the invocation of an asycnhornous task returns a future object that represents a dependency on the completion of that task that can be waiting or used to create a continuation task. This allows tasks to be linked together by their dependencies on each other.
```
AerialMantis: So this proposal extends the idea of continuations as a way of chaining execution of functions to incorporate the movement of data. This way users or library developers who wish to use a continuation style interface can also use to describe their data dependencies and chain them together with computations.

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -5,222 +5,384 @@
...
 
-The **execution context** must encapsulate both the execution agents and a global memory region that is accessible from those execution agents. It is required that the memory region of the **execution context** be cache coherent across the execution agents created from the same invocation of an execution function. It is not required that the memory region of the **execution context** be cache coherent across execution agents created from different execution functions or on different **execution contexts**.
+Another data flow model is one that uses a runtime to create a graph of tasks, which can be optimised.
 
-*Note: some systems are capable of providing cache coherency across execution functions or **execution contexts** through the availability of shared virtual memory addressing or shared physical memory. This feature is discussed in more detail in the future work section.*
+Another data flow model is one which simply submits work to be executed and waits for it all to complete, typically called fire and forget execution.
```
AerialMantis: This was more of a special case of another more general data-flow model, it was not a good example so I have removed it.

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -5,222 +5,384 @@
...
+### 2.5. Must be interoperable between executor implementations
 
-Synchronisation operations are course grained synchronisation in that they synchronise the entire managed memory allocation of a ``managed_ptr``.
+The unified interface for execution will have may different implementation which target different platforms and therefore a unified interfacec for data movement would too require this range of implementations.
 
-*Note: some systems are capable of providing finer-grained synchronisation via atomic operations, however, this is generally only when there is shared virtual memory addressing or shared physical memory. This feature is discussed in more detail in the future work section.*
+It's important therefore that a unified interface for data movement be interoperable between different executor implementations, it must be possible to synchronise data between the memory region of one executor implementation and another.
```
AerialMantis: You raise a good question here, this is something which perhaps needs to be better described. So this proposal will require each implementation of an executor to provide implementations of the synchronization functions `put` and `get` for any host based device which they support synchronization from (this will generally be a host CPU but does not have to be). These functions will perform an implementation defined set of commands to make a `managed_ptr` available on the memory region of an executor or on the memory region of the host respectively. When synchronizing directly between two executors of different implementations these will have to be synchronized via a host by calling `get` on the source executor and `put` on the destination executor. This is the default behavior of the `sync` function, however this can be optimized by providing an implementation defined channel which is specialized for synchronizing between those executors.

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -5,222 +5,384 @@
...
 
-}  // namespace execution
-}  // namespace experimental
-}  // namespace std
-```
+In order to support a wide range of systems, both heterogneous and distributed, a high level absraction for the `managed_ptr` is nessesary, however not all devices will be able to support of all of a `managed_ptr`s capabilities.
+
+As the `managed_ptr` can be passed between and used on different devices it can therefore have different capabilities depending on where it is currently being used. The `managed_ptr` will likely be compiled separately or a least have different restrictions imposed on the code that is executed depending on where it s used. It is therefore important to be able to determine what capabilities the `managed_ptr` supports where it is being used.
+
+The major distinction is whether or not the `executor_type` of a `managed_ptr` is a *host* which or a *remote*. If the `executor_type` is a *host* then it supports host based capabilities; being able to perform allocation and sycnhronisation operations, if it is a *remote* it does not. Whether the `executor_type` of a `managed_ptr` is a *host* can be queried with the static constexpr bool `is_host_based`.
```
AerialMantis: Fixed.

AerialMantis: The term `host` refers to any device which is capable of performing synchronization and allocation operations and the term `remote` refers to any device which cannot. Other devices on a system in this sense can be a `host` as they too can execute functions and perform synchronization operations.

This current distinction of `host` and `remote` is rather limited, it does not take into consideration the full range of device capabilities. However as these capabilities must come from the executor and therefore are also a consideration for the executors proposal, this has been left out of the scope of this paper for the time being.

Though this is an important future work.

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -5,222 +5,384 @@
...
+
+### 1.2. Motivation
+
+In non-heterogeneous or non-distributed systems, there is typically a single device: a CPU with a single memory region or an area of addressable memory. In contrast, heterogeneous and distributed systems have multiple devices with shared or discrete memory regions.
+
+A device in this respect can be any architecture that exists within a system that is C\+\+ programmable; this can include CPUs, GPGPUs, APUs, FPGAs, DSPs, NUMA nodes, I/O devices and other forms of accelerators.
 
 This introduces additional complexity for accessing a memory allocation that is not present in current C\+\+, the requirement that said memory allocation be available in one of many memory regions within a system throughout a program's execution. For the purposes of this paper, we will refer to such a memory allocation as a managed memory allocation as it describes a memory allocation that is accessible consistently across multiple memory regions. With this requirement comes the possibility that a memory region on a given device may not have the most recently modified copy of a managed memory allocation, therefore requiring synchronisation to move the data.
```
AerialMantis: Done.

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -5,222 +5,384 @@
...
 
 As the act of dispatching work to a remote device to be executed was once only a problem for third party APIs and so too was the act of moving data to those devices for computations to be performed on. However, now that C\+\+ is progressing towards a unified interface for execution [1] the act of moving data to remote devices to be accessible for dispatch via executors is now a problem for C\+\+ to solve as well, a unified interface for data movement of such. The act of moving data to a remote device is very tightly coupled with the work being performed on said remote device. This means that this unified interface for data movement must also be tightly coupled with the unified interface for execution.
```
AerialMantis: Fixed.

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -5,222 +5,384 @@
...
 namespace experimental {
+inline namespace concurrency_v2 {
 namespace execution {
 
+/* extensions to executor classes */
+class executor {
+  ...
+
+  /* aliases */
+  using global_memory_allocator_type = __global_memory_allocator_type__;
```
AerialMantis: Tbh I wasn't sure about this name, though I couldn't think of a better term to describe it. This is the allocator for allocating memory within the executor's memory region. Perhaps `executor_memory_allocator_type` would be better.

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -5,222 +5,384 @@
...
 
-The ``managed_ptr`` class template has a single template parameter; ``T`` specifying the type of the elements.
+In order to facilitate the `managed_ptr`, it is necessary to introduce requirements on the *executor* and the *execution context*; the class associated with an *executor* which encapsulates the underlying execution resource on which *execution agents* are executed.
 
-The type parameter ``T`` specifies the element type of the container, any ``T`` satisfies the ``managed_ptr`` element type requirements if:
+#### 3.2.1. Execution Context Memory Region
 
-* ``T`` is a standard layout type.
-* ``T`` is copy constructible.
+In addition to a number of *execution agents* an *execution context* must also encapsulate at least one memory region that is concurrently accessible by all *execution agents* of any single invcation of an execution function. It is not required that the memory region be concurrently accessible to *execution agents* of another invocation of an execution function or *execution agents* executing on a different *execution context*.
```
AerialMantis: Fixed.

AerialMantis: Yes this is intentional, if a system has a fine grained unified address space it is possible for two execution agents in different invocations to access the same memory region concurrently. However the `managed_ptr` in it's current state does not permit this and therefore does not require it.

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -5,222 +5,384 @@
...
 
-A  ``managed_ptr`` can be constructed in a number of ways:
+> This requirement currently satisfies the course grained level of synchronisation described in requirement 2. These requirements will be loosened in a later iteration of this paper when a suitable interface for fine grained sycnrhonisation is included. The curent proposal for this is described in section 5.1 as future work.
```
AerialMantis: Fixed.

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -5,222 +5,384 @@
...
 
-* for constructor (1) the  ``managed_ptr`` allocates the number of elements specified by the `size_t` parameter using the default allocator, therefore allocating on the host remote device memory region.
-* for constructor (2) the  ``managed_ptr`` takes ownership of the pointer parameter.
-* for constructor (3) the  ``managed_ptr`` takes ownership of the const_pointer parameter.
-* for constructor (4) the  ``managed_ptr`` allocates the number of elements specified by the `size_t` parameter using the allocator specified by the allocatorT parameter.
+#### 3.2.2. Executor Global Memory Allocator
 
-The default constructor T is not called for the elements of the ``managed_ptr`` during allocation.
+The *execution context* must also provide an alias to an allocator type called `global_memory_allocator_type` that is capable of allocating memory in it's memory region, and a member function for retrieving an instance of this type called `global_memory_allocator()`.
```
AerialMantis: This has been changed to `executor_memory_allocator_type`.

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -5,222 +5,384 @@
...
 
-The destructor is not required to perform any synchronisation operations.
+### 3.3. Managed Pointer Class Template
 
-The member function ``get()`` triggers a synchronisation operation to the host CPU if the host CPU returns a ``future_type`` object which can be used to wait on the operation completing. If an error occurred during the synchronisation point or the pointer to the accessible **execution context** is no longer valid then an exception is thrown and stored within the ``future_type`` that is returned. It is undefined to call ``get()`` within a function executing on an **execution context**.
+The proposed addition to the standard library is the `managed_ptr` class template. The `managed_ptr` is a smart pointer which has ownership of a contiguous managed allocation of memory that can be shared between any number of devices  represented by their respective *execution contexts*.
```
AerialMantis: Fixed.

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -5,222 +5,384 @@
...
+The type parameter `T` specifies the pointer type, any `T` satisfies the `managed_ptr` pointer type requirements if:
 
-The member function ``size`` returns the number of elements of type ``T`` stored within the managed allocation.
+> Another solution for read-only access was considered where access to a `managed_ptr` instance could be dynamically specified via properties using the proposed `array_ref` interface [10]. However it was felt that this approach would be too verbose and would introduce unnessesary overhead for casess which did not require the properties to be specified. Additionally as many implementation would wish to make use of read-only partitions of memory this could require an implementation to copy in an out of this memory partition dynamically which could introduce opaque latency in synchronisation.
```
AerialMantis: Fixed.

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -5,222 +5,384 @@
...
-executor_future_t<Executor, void> then_get(Predicate pred, Executor, ManagedPtrTN...);
+#### 3.3.2. Execution context capabilities
 
-}  // namespace execution
-}  // namespace experimental
-}  // namespace std
-```
+In order to support a wide range of systems, both heterogneous and distributed, a high level absraction for the `managed_ptr` is nessesary, however not all devices will be able to support of all of a `managed_ptr`s capabilities.
+
+As the `managed_ptr` can be passed between and used on different devices it can therefore have different capabilities depending on where it is currently being used. The `managed_ptr` will likely be compiled separately or a least have different restrictions imposed on the code that is executed depending on where it s used. It is therefore important to be able to determine what capabilities the `managed_ptr` supports where it is being used.
```
AerialMantis: Fixed.

managed-pointer/cpp-20/D0567R1.md
AerialMantis: After thinking about this further I think there is a way in which we can describe the properties of an executor with respect to data movement in a way which is fine grained and extensible.

I have updated the paper to specify traits which can be used to query whether not an executor has a particular capability, rather than grouping them into the `host` and `remote` categories:

```cpp
template<class T> constexpr bool can_put_v = can_put<T>::value;
template<class T> constexpr bool can_get_v = can_get<T>::value;
template<class T> constexpr bool can_sync_v = can_sync<T>::value;
template<class T> constexpr bool can_allocate_v = can_allocate<T>::value;
template<class T> constexpr bool can_deallocate_v = can_deallocate<T>::value;
```

managed-pointer/cpp-20/D0567R1.md
```diff
@@ -5,222 +5,384 @@
...
+
+The major distinction is whether or not the `executor_type` of a `managed_ptr` is a *host* which or a *remote*. If the `executor_type` is a *host* then it supports host based capabilities; being able to perform allocation and sycnhronisation operations, if it is a *remote* it does not. Whether the `executor_type` of a `managed_ptr` is a *host* can be queried with the static constexpr bool `is_host_based`.
+
+> For the purposes of this paper the capabilities of host and remote execution contexts simply state that allocation and sycnhronisation operations can only be performed by host execution contexts. In reality the capabilities of execution contexts will be far more complex and will require an depth interface for querying said capabilities. However this problem of language capabilities on execution contexts particularly for heterogeneous architectures is a much larger problem for C\+\+ that goes beyond the scope of this paper and indeed the executors proposal as well.
```
AerialMantis: See previous comment.

managed-pointer/cpp-20/D0567R1.md
AerialMantis: Fixed.

managed-pointer/cpp-20/D0567R1.md
AerialMantis: Fixed.

managed-pointer/cpp-20/D0567R1.md
AerialMantis: Fixed.

managed-pointer/cpp-20/D0567R1.md
AerialMantis: The reason for maintaining the `put` and `get` functions as well as `sync` is that all of these functions take a variadic pack of `managed_ptr` as a parameter and the host executor could potentially be different for each. Though it's most common that they would be the same.

We could perhaps simplify the interface to just the `sync` function:

```cpp
  template <typename SynchronizationChannel, typename SrcExecutor = managed_ptr<dataT>::host_executor_type,
            typename DestExecutor, typename dataT>
  future_type_t<void>
  sync(SynchronizationChannel = __default_synchronization_channel__{},
       SrcExecutor, DestExecutor, managed_ptr<dataT> managedPtr);
```

Though this would mean that only one managed_ptr could be passed to `sync` at a time and would perhaps  reduce the clarity of the semantics.

managed-pointer/cpp-20/D0567R1.md
AerialMantis: Fixed.

managed-pointer/cpp-20/D0567R1.md
AerialMantis: Fixed.

managed-pointer/cpp-20/D0567R1.md
AerialMantis: Fixed.

managed-pointer/cpp-20/D0567R1.md
AerialMantis: Fixed.

AerialMantis: There is a requirement that the `managed_ptr` does not call the constructor of the object it is managing, this is specified earlier in the paper. The reason for this is that the `managed_ptr` is either constructed from a pointer in which case the object has already been constructed or it is constructed from an allocator in which case it is not always possible for an executor's allocator to call the constructor on the alllocated memory.

managed-pointer/cpp-20/D0567R1.md
AerialMantis: If you were to pass a `managed_ptr` to a function being executed on an executor on which the `managed_ptr` is not accessible that would be regarded as undefined behavior. It's the users responsibility to ensure the `managed_ptr`is accessible on an executor before passing it to a function executing on it.

managed-pointer/cpp-20/D0567R1.md
AerialMantis: Yes, so the `then_get` function is essentially the `get` function with a future predicate, if it is not already accessible, it will synchronize the managed memory allocation with to host executor.

managed-pointer/cpp-20/D0567R1.md
AerialMantis: Fixed.

managed-pointer/cpp-20/D0567R1.md
AerialMantis: Fixed.

managed-pointer/cpp-20/D0567R1.md
AerialMantis: This is the question that will need to be answered. On the one hand you could say no because if you have a system with a unified address space then you can likely just pass a raw pointer directly. However if you wish to inter-operate with executors in systems which don't, or if you want to write generic code to support a range of different systems which may nor may not have a unified address space, or if you want manage the ownership of pointers between executors then you may still want to use a `managed_ptr`.

managed-pointer/cpp-20/D0567R1.md
AerialMantis: I meant here that this idea has not been integrated into the proposal yet. I've re-worded this.

managed-pointer/cpp-20/D0567R1.md
AerialMantis: Fixed.

managed-pointer/cpp-20/D0567R1.md
AerialMantis: This is a possibility that we should consider.

managed-pointer/cpp-20/D0567R1.md
AerialMantis: Fixed.

PR #9, opened 2017-06-08
Updated buffer properties proposal

Ruyk:
The buffer tags are now buffer properties.
Buffer properties can take parameters, which simplifies the interface
and enables adding extensions.

The tied-buffer proposal has been refurbished to work with properties
instead of derived classes.

### DIFF
--- a/tied-buffer/sycl-2.2/buffer-tags.md
+++ b/tied-buffer/sycl-2.2/buffer-tags.md
@@ -1,8 +1,8 @@
-# Buffer tied to a single context
+# Buffer Properties
 
 | Proposal ID | CP008 |
 |-------------|--------|
-| Name | Buffer tied to a single context |
+| Name | Buffer Properties |
 | Date of Creation | 14 March 2017 |
 | Target | SYCL 2.2 |
 | Current Status | _Work In Progress_ |
@@ -12,7 +12,13 @@
 
 ## Overview
 
-This proposal aims to define an interface for specialising the construction of the buffer class whilst still allowing buffers of different types to be stored in a generic container, via the use of buffer tags.
+This proposal aims to define an interface for specialising the construction of the buffer class whilst still allowing buffers of different types to be stored in a generic container, via the use of buffer tags. 
+Each tag represent a different property of a buffer.
+
+## Revisions
+
+This revision clarifies the behaviour of the tags and how they map to buffer
+properties.
 
 ## Requirements
 
@@ -54,14 +60,15 @@ class buffer {
 With this interface, a user can specify a specialisation of the buffer class by simply providing one or more values of the following tag enum class to the constructor of a buffer.
 
 ```cpp
-enum class tag {
-  context_bound,
-  mapping,
-  svm
+namespace property {
+  struct map_ptr;
+  struct cl_interop;
+  struct gl_interop;
+  struct svm;  // SYCL 2.2 Only
 };
 ```
 
-Some of these tags will require an additional interface for the buffer class, for example, the `tag::context_bound` tag will need a `get_context()` member function.
+Some of these tags will require an additional interface for the buffer class, for example, the `property::context_bound` tag will need a `get_context()` member function.
 
 ```cpp
 template <typename dataT, int kElems, typename allocatorT = buffer_allocator<dataT>>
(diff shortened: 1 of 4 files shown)


tied-buffer/sycl-2.2/buffer-tags.md
```diff
@@ -74,36 +81,89 @@ class buffer {
...
+};
+```
+
+* *cl_interop*: Buffer constructed for interoperability with OpenCL cl\_mem objects
 
+```cpp
+struct cl_interop {
   ...
+  // Construct a property for interoperability
+  cl_interop(cl_mem clMemObject) { }
```
Ruyk: this should also take an event and a sycl queue

tied-buffer/sycl-2.2/buffer-tags.md
```diff
@@ -14,6 +14,9 @@
 
 This proposal aims to define an interface for specialising the construction of the buffer class whilst still allowing buffers of different types to be stored in a generic container, via the use of buffer tags. 
 Each tag represent a different property of a buffer.
+By defining properties for buffers and enabling expressing them independently from the
```
AerialMantis: `enabling expressing them` -> `allowing them to be expressed`

tied-buffer/sycl-2.2/buffer-tags.md
```diff
@@ -12,7 +12,16 @@
 
 ## Overview
 
-This proposal aims to define an interface for specialising the construction of the buffer class whilst still allowing buffers of different types to be stored in a generic container, via the use of buffer tags.
+This proposal aims to define an interface for specialising the construction of the buffer class whilst still allowing buffers of different types to be stored in a generic container, via the use of buffer tags. 
+Each tag represent a different property of a buffer.
+By defining properties for buffers and enabling expressing them independently from the
```
AerialMantis: `enabling expressing them` -> `allowing them to be expressed`

tied-buffer/sycl-2.2/buffer-tags.md
```diff
@@ -54,14 +63,15 @@ class buffer {
...
+namespace property {
+  struct map_ptr;
+  struct cl_interop;
+  struct gl_interop;
+  struct svm;  // SYCL 2.2 Only
 };
 ```
 
-Some of these tags will require an additional interface for the buffer class, for example, the `tag::context_bound` tag will need a `get_context()` member function.
+Some of these tags will require an additional interface for the buffer class, for example, the `property::context_bound` tag will need a `get_context()` member function.
```
AerialMantis: If we are requiring that properties be passes as an instance, for example `property::context_bound(myContext)`, we could then potentially extend the `buffer` interface in a generic and extensible way through these objects. For example if you wanted to get the context of a context bound buffer you could do something like `buf.get_property<property::context_bound>().get_context()`.

This would remove any requirement to extend the `buffer` interface and handle exceptions for invalid uses of it.

AerialMantis: This would require an additional member function for retrieving the property `template <typename propertyT> propertyT get_property()`.

tied-buffer/sycl-2.2/buffer-tags.md
```diff
@@ -74,41 +84,222 @@ class buffer {
...
+  map_ptr() = default;
+
+  // Implementation defined members
+};
+```
+
+* *cl_interop*: Buffer constructed for interoperability with OpenCL cl\_mem objects
+
+```cpp
+struct cl_interop {
```
AerialMantis: I'm not sure about having OpenCL interop as a property as the interop feature is standardized across the SYCL API, so this may cause some confusion if the interop API for buffers is different from everything else.

tied-buffer/sycl-2.2/buffer-tags.md
```diff
@@ -74,41 +84,222 @@ class buffer {
...
 ```
 
+## Defined properties
+
+### Extension to SYCL 1.2
+
+* *map_ptr*: Buffer will use the given pointer exclusively for host access.
+
+```cpp
+struct map_ptr {
```
AerialMantis: The name `map_ptr` differs from the name used later `property::map`.

tied-buffer/sycl-2.2/buffer-tags.md
Ruyk: fixed

tied-buffer/sycl-2.2/buffer-tags.md
Ruyk: That means the buffer class must construct a new instance of the property with the required values (since we cannot mandate storing the original instance the user created), can you prototype that?

tied-buffer/sycl-2.2/buffer-tags.md
Ruyk: fixed

tied-buffer/sycl-2.2/buffer-tags.md
Ruyk: The buffer/images are special cases, since according to the specification OpenCL interop buffers are technically tied to the original context. This modifies the normal behaviour of the buffer, which seems to justify a different property. In other SYCL objects with interop, the normal behaviour of the object is preserved (e.g. context or kernel interop)

tied-buffer/sycl-2.2/buffer-tags.md
AerialMantis: yes that's right, though in most cases the property object will be very lightweight so shouldn't incur any significant overhead.

We would also have to require that calling the `get_property` member function with a property type that the buffer does not have, throws an exception. Which should be reasonable as we also have the `has_property` member function to query if the buffer does have the property.

Yes, I can prototype this.

Ruyk: Can we use http://en.cppreference.com/w/cpp/utility/optional for get_property?

PR #13, opened 2017-06-16
Extends with example in TensorFlow

lukeiwanski:


Ruyk:
Looks good to me, thanks for the link to the TF issue. Just to clarify: Most of the line ending changes were intentional?

lukeiwanski:
Kind of. Apparently atom removes last white space even from the markdown files.
But then, I assumed they were not intended to be there in the first place. I will revert if that was intentional.

Ruyk:
Not really, just curious. Seems Vim does the opposite :smile: 
Thanks! I am merging changes in.

PR #15, opened 2017-06-30
Fix #14: Added fill method to the handler

### LINKED ISSUE #14: CP001: Missing SYCL memset capabilities
Adding memset to SYCL by using https://www.khronos.org/registry/OpenCL/sdk/1.2/docs/man/xhtml/clEnqueueFillBuffer.html

Ruyk:
Adds a method to the handler class to fill a memory object with a certain value


AerialMantis:
Looks good to me :+1:

Issue #16, opened 2017-07-04
CP008: Add mutex buffer property


AerialMantis:
With the new proposal for buffer properties the optional mutex parameter of the buffer constructor can instead be provided as a property. This will reduce the number of buffer constructors and allow the mutex to be used in combination with other constructors not possible before.

PR #17, opened 2017-07-04
Issue #16: Add mutex property to buffer properties proposal.

AerialMantis:
* Add mutex property to list of properties and add example.


tied-buffer/sycl-2.2/buffer-tags.md
```diff
@@ -75,6 +75,7 @@ namespace buffer {
   struct use_host_ptr;
   struct cl_interop;
   struct svm;  // SYCL 2.2 Only
+  struct mutex { mutex(mutex_class &mutexRef); };
```
Ruyk: can you change the name of the property to "use_mutex" instead ? it avoids confusions with the C++ names

AerialMantis: Done.

PR #18, opened 2017-07-10
Placeholder accessors can be default constructed

Ruyk:
* Placeholder accessor constructor can optionally not take memobj

* Handler can specify the buffer later on

### DIFF
--- a/placeholder_accessors/index.md
+++ b/placeholder_accessors/index.md
@@ -9,7 +9,7 @@
 | Reply-to | Ruyman Reyes <ruyman@codeplay.com> |
 | Original author | Ruyman Reyes <ruyman@codeplay.com> |
 | Requirements | CP001 |
-| Contributors | Gordon Brown <gordon@codeplay.com>, Victor Lomuller <victor@codeplay.com>, Mehdi Goli <mehdi.goli@codeplay.com>, Peter Zuzek <peter@codeplay.com>, Luke Iwanski <luke@codeplay.com> |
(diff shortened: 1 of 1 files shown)


placeholder_accessors/index.md
```diff
@@ -192,6 +195,12 @@ The handler gains a new method,
 that registers the requirements of the placeholder accessor on the given
 command group.
 
+Another method, that allows specifying the memory object the placeholder
+accessor will be associated is also provided:
+
+`handler::require(buffer<T, dim> b,
```
Ruyk: The has_buffer is on the prototype but not on the proposal, good catch!

placeholder_accessors/index.md
```diff
@@ -192,6 +195,12 @@ The handler gains a new method,
 that registers the requirements of the placeholder accessor on the given
```
AerialMantis: If we wish to support the option for constructing a regular accessor from a placeholder accessor (i.e. `get_access`) then this interface should also be extended to allow you to specify a memory object. We should also specify that an exception should be thrown if `get_access` is called without providing a memory object on a placeholder accessor which doesn't have a memory object already.

On a side note, perhaps we should consider changing the naming of this member function, `get_access` could be confused with the regular `get_access` member functions.

ProGTX:
I believe https://github.com/codeplaysoftware/standards-proposals/pull/124 offers a suitable alternative to this.

PR #20, opened 2017-07-19
Fill pattern size restrictions

ProGTX:
In the documentation for [`clEnqueueFillBuffer` ](https://www.khronos.org/registry/OpenCL/sdk/1.2/docs/man/xhtml/clEnqueueFillBuffer.html), the pattern size needs to be one of 1, 2, 4, 8, 16, 32, 64, 128. It makes sense to have this restriction for the `fill` function as well because enabling it in the general case would be very inefficient.

Also fixed an accessor name typo.

### DIFF
--- a/asynchronous-data-flow/sycl-2.2/03_interacting_with_data_on_the_host.md
+++ b/asynchronous-data-flow/sycl-2.2/03_interacting_with_data_on_the_host.md
@@ -151,9 +151,10 @@ The example shows how to fill buffer A contents with an scalar.
 Note that at least write access is required.
 
 ```cpp
+int pattern = 10;
 auto cgH = [=] (handler& h) {
   auto accA = bufA.get_access<access::mode::write>(h);
-  h.fill(accA, 10);
+  h.fill(accA, pattern);
 };
 qA.submit(cgH);
 ```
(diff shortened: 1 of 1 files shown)


asynchronous-data-flow/sycl-2.2/03_interacting_with_data_on_the_host.md
```diff
@@ -153,11 +153,15 @@ Note that at least write access is required.
 ```cpp
 auto cgH = [=] (handler& h) {
   auto accA = bufA.get_access<access::mode::write>(h);
-  h.fill(accB, 10);
+  h.fill(accA, 10);
 };
 qA.submit(cgH);
 ```
 
+Because of device restrictions on the size of the pattern, the size in bytes of
```
Ruyk: Because of restrictions in OpenCL implementations

asynchronous-data-flow/sycl-2.2/03_interacting_with_data_on_the_host.md
```diff
@@ -153,11 +153,15 @@ Note that at least write access is required.
 ```cpp
 auto cgH = [=] (handler& h) {
   auto accA = bufA.get_access<access::mode::write>(h);
-  h.fill(accB, 10);
+  h.fill(accA, 10);
```
Ruyk: Can you change the example to use a variable instead of a literal? just to make it a bit more ellaborate

ProGTX:
Addressed comments

PR #23, opened 2018-01-22
SVM proposal

ProGTX:
Proposal to simplify Shared Virtual Memory in SYCL 2.2.


Ruyk:
Looks generally good, we can take this as a base and use other pull request to amend and improve things

PR #27, opened 2018-02-07
CP013: Apply proposals for relative affinity and allocation.

AerialMantis:
* Introduce the affinity_query class, with operators.
* Add the execution_context::memory_resource() member function.
* Change this_system::resource() to this_system::resources() and have it
return a vector of execution_resource instead.
* Add requirements on this_system::resources().
* Add capability to migrate memory and work between resources to list of
problem space challenges.
* Add minor reformatting to proposed wording.


affinity/cpp-20/d0796r1.md
```diff
@@ -136,25 +142,54 @@ There are some additional challenges which we have been investigating but are no
 
     /* Execution context */
 
-    struct execution_context {
+    class execution_context {
+     public:
 
       using executor_type = __unspecfied__;
```
Ruyk: unspecfied -> unespecified

affinity/cpp-20/d0796r1.md
```diff
@@ -167,55 +202,56 @@ There are some additional challenges which we have been investigating but are no
 
 The `execution_resource` class provides an abstraction over a system's hardware capable of memory allocation, execution of light weight exeution agents or both.
 
-### `execution_resource` constructors
+> [*Note:* The `execution_resource` is required to be implemented such that the underlying software abstraction is initialised on when the `execution_resource` is constructed, maintained through reference counting and cleaned up on destruction of the final reference. *--end note*]
```
Ruyk: Do we really need to mandate an implementation detail like this? if your execution resource is just a thread id for example, I don't think you need reference counting

affinity/cpp-20/d0796r1.md
```diff
@@ -225,35 +261,83 @@ The `execution_context` class provides an abstraction for managing a number of l
...
 
 *Returns:*
 
-	executor_type executor() noexcept;
+    executor_type executor() noexcept;
 
 *Returns:*
 
-## Free functions
+    pmr::memory_resource *memory_resource() noexcept;
```
Ruyk: see my comments above

affinity/cpp-20/d0796r1.md
```diff
@@ -225,35 +261,83 @@ The `execution_context` class provides an abstraction for managing a number of l
...
 
+> [*Note:* The comparison operators rely on the availability of the `expected` class template (see [P0323r4: std::expected][p0323r4]), if this does not become available then an alternative error/value construct will be adopted instead. *--end note*]
+
+## Free functions
+
+	std::vector<execution_resource> resources() noexcept;
+
+*Returns:* A std::vector containing all system level resources.
+
+*Requires:* If `resources().size() > 0`, `resources()[0]` be the `execution_resource` corroponding to the current thread of execution. The value returned by `resources()` be the same at any point after the invocation of `main`.
```
Ruyk: I think we should require resources().size() to be at least 1.

affinity/cpp-20/d0796r1.md
```diff
@@ -33,7 +37,7 @@ One strategy to improve applications' performance, given the importance of affin
...
 
-The affinity problem is especially challenging for applications whose behavior changes over time or is hard to predict, or when different applications interfere with each other's performance. Today, most OSes already can group processing units according to their locality and distribute processes, while keeping threads close to the initial thread, or even avoid migrating threads and maintain first touch policy. Nevertheless, most programs can change their work distribution, especially in the presence of nested parallelism.
+The affinity problem is especially challenging for applications whose behavior changes over time or is hard to predict, or when different applications interfere with each other's performance. Today, most OSes already can group processing units according to their locality and distribute processes, while keeping threads close to the initial thread, or even avoid migrating threads and maintain first touch policy. Nevertheless, most pro  grams can change their work distribution, especially in the presence of nested parallelism.
```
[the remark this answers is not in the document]
AerialMantis: Fixed.

affinity/cpp-20/d0796r1.md
```diff
@@ -93,6 +97,7 @@ In this paper we describe the problem space of affinity for C++, the various cha
 * How to represent, identify and navigate the topology of execution resources available within a heterogeneous or distributed system.
 * How to query and measure the relative affininty between different execution resources within a system.
 * How to bind execution and allocation particular execution resource(s).
+* How to migrate memory work and memory allocations between execution resources.
```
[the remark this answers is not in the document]
AerialMantis: So this isn't covered by our proposed solution, it'll be in the future work, but I think we should still list it in the background research.

affinity/cpp-20/d0796r1.md
```diff
@@ -136,25 +142,54 @@ There are some additional challenges which we have been investigating but are no
...
+      friend expected<size_t, error_type> operator<=(const affinity_query&, const affinity_query&);
+      friend expected<size_t, error_type> operator>=(const affinity_query&, const affinity_query&);
+
     };
 
     /* This system */
 
     namespace this_system {
-      const execution_resource &resource();
+      std::vector<execution_resource> resources() noexcept;
```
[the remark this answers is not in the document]
AerialMantis: My expectation with the current approach would be that if a dynamic library introduces a new device at runtime it would not affect the topology as seen by `this_system::resources()`.

affinity/cpp-20/d0796r1.md
```diff
@@ -136,25 +142,54 @@ There are some additional challenges which we have been investigating but are no
...
-      execution_context(ExecutionResource &&execResource);
+      execution_context(const execution_resource &);
 
       ~execution_context();
 
       const execution_resource &resource() const noexcept;
 
       executor_type executor() noexcept;
 
+      pmr::memory_resource *memory_resource() noexcept;
```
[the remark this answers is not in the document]
AerialMantis: So originally made this by pointer as the `pmr::polymorphic_allocator` expects a raw pointer to a memory resource, but I agree that this is not ideal. I think this should be returned by reference instead since if it's returned by value you'll end up with a copy of the memory resource.

You make a good point about the `noexcept`, I will remove this. I'll also put a throws clause to state that an exception will be thrown if the execution context cannot allocate memory. And i'll add a similar requirement for the `executor` member function.

I'll add an `allocator` member function too.

affinity/cpp-20/d0796r1.md
AerialMantis: Done.

affinity/cpp-20/d0796r1.md
AerialMantis: This is what we require now, or do you mean we should say `>= 1` rather than `> 0`?

affinity/cpp-20/d0796r1.md
AerialMantis: Tbh this requirement was quite hard to word, so it could definitely be improved. The reason I put this is that, since we return `execution_resource`s by value, they need to be reference counted to guarantee that two `execution_resource`s which identify the same hardware resource reference the same internal implementation. For example `execution_context(resource).resource() == resource`. Though this gets into equality semantics which are not yet defining.

affinity/cpp-20/d0796r1.md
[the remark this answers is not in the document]
Ruyk: The sentence says what we describe in the paper, which I think we don't. I suggest we just remove this for now, we can add it later.

affinity/cpp-20/d0796r1.md
```diff
@@ -112,20 +117,21 @@ There are some additional challenges which we have been investigating but are no
...
+      execution_resource &operator=(const execution_resource &);
+      execution_resource &operator=(execution_resource &&);
       ~execution_resource();
 
       size_t concurrency() const noexcept;
-      size_t partition_size() const noexcept;
 
-      const execution_resource &partition(size_t i) const noexcept;
-      const execution_resource &member_of() const noexcept;
+      std::vector<resource> resources() const noexcept;
```
Ruyk: define an alias for this in a single place. We'll likely have to change this later anw

affinity/cpp-20/d0796r1.md
```diff
@@ -136,25 +142,60 @@ There are some additional challenges which we have been investigating but are no
...
+      execution_context(const execution_resource &) noexcept;
 
       ~execution_context();
 
       const execution_resource &resource() const noexcept;
 
-      executor_type executor() noexcept;
+      executor_type executor() const;
+
+      pmr_memory_resource_type &memory_resource() const;
```
Ruyk: why a reference?

affinity/cpp-20/d0796r1.md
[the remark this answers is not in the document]
Ruyk: I think a dynamic library should be able to populate the system resources before the first call (notionally before main, but I guess is easier to define before the first call to resources). Otherwise how the dynamic loading of OpenCL would load the icd, for example?

affinity/cpp-20/d0796r1.md
```diff
@@ -165,94 +206,163 @@ There are some additional challenges which we have been investigating but are no
 
 ## Class `execution_resource`
 
-The `execution_resource` class provides an abstraction over a system's hardware capable of memory allocation, execution of light weight exeution agents or both.
+The `execution_resource` class provides an abstraction over a system's hardware capable of memory allocation, execution of light weight exeution agents or both. An `execution_resource` can represent further `execution_resource`s, these `execution_resource`s are said to be *members of* this `execution_resource`.
```
Ruyk: exeution -> execution
"can represent further" -> can contain other

affinity/cpp-20/d0796r1.md
```diff
@@ -165,94 +206,163 @@ There are some additional challenges which we have been investigating but are no
...
 
-### `execution_resource` constructors
+> [*Note:* The `execution_resource` is required to be implemented such that the underlying software abstraction is initialised on when the `execution_resource` is constructed, maintained through reference counting and cleaned up on destruction of the final reference. *--end note*]
```
Ruyk: on when -> when
why the reference counting?

affinity/cpp-20/d0796r1.md
```diff
@@ -165,94 +206,163 @@ There are some additional challenges which we have been investigating but are no
...
-	~execution_resource() = delete;
+    ~execution_resource();
 
 ### `execution_resource` operations
 
-      size_t concurrency() const noexcept;
+    size_t concurrency() const noexcept;
 
-*Returns:*
+*Returns:* The collective concurrency available to this resource. More pecifically, the number of *threads of execution* collectively available to this `execution_resource` and any resources which are *members of*, recursively.
```
Ruyk: collective concurrency -> total concurrency
pecifically -> specifically

affinity/cpp-20/d0796r1.md
```diff
@@ -165,94 +206,163 @@ There are some additional challenges which we have been investigating but are no
...
 
-      size_t partition_size() const noexcept;
+    std::vector<resource> resources() const noexcept;
 
-*Returns:*
+*Returns:* All `execution_resource`s which are *members of* this resource.
 
-      const execution_resource &partition(size_t i) const noexcept;
+    const execution_resource &member_of() const noexcept;
```
Ruyk: why a reference?

affinity/cpp-20/d0796r1.md
```diff
@@ -165,94 +206,163 @@ There are some additional challenges which we have been investigating but are no
...
+    std::string name() const noexcept;
 
-*Returns:*
+*Returns:* An implementation defined string.
 
-      std::string name() const noexcept;
+    bool can_place_memory() const noexcept;
 
-*Returns:*
+*Returns:* If this resource is capable of allocating memory with affinity, 'true'.
```
Ruyk: remove "with affinity"

affinity/cpp-20/d0796r1.md
```diff
@@ -165,94 +206,163 @@ There are some additional challenges which we have been investigating but are no
...
+    execution_context(const execution_resource &) noexcept;
 
-### `execution_context` destructor
+*Effects:* Constructs an `execution_context` with the provided resource as the *contained resource*.
 
+### `execution_context` destructor
+    
     ~execution_context();
 
+*Effects:* May or may not block to wait any work bveing executed on the *contained resource*.
```
Ruyk: bveing -> being

Ruyk: this should be clarified, destructor should wait, otherwise we risk leaking resources after main finishes and messing up with dynamic libraries

affinity/cpp-20/d0796r1.md
```diff
@@ -165,94 +206,163 @@ There are some additional challenges which we have been investigating but are no
...
+*Returns:* All `execution_resource`s which are *members of* this resource.
 
-      const execution_resource &partition(size_t i) const noexcept;
+    const execution_resource &member_of() const noexcept;
 
-*Returns:*
+*Returns:* The `execution_resource` which this resource is a *member of*.
 
-      const execution_resource &member_of() const noexcept;
+    std::string name() const noexcept;
```
Ruyk: can this return a string and be noexcept at the same time? what is the expected value of std::string if the system cannot allocate memory for the string for example? are you expecting a terminate?

affinity/cpp-20/d0796r1.md
```diff
@@ -165,94 +206,163 @@ There are some additional challenges which we have been investigating but are no
...
+> [*Note:* An affinity query is permitted to fail if affinity between the two execution resources cannot be calculated for any reason, such as the resources are of different vendors or communication between the resources is not possible. *--end note*]
+
+> [*Note:* The comparison operators rely on the availability of the `expected` class template (see [P0323r4: std::expected][p0323r4]), if this does not become available then an alternative error/value construct will be adopted instead. *--end note*]
 
 ## Free functions
 
-	const this_system::execution_resource &resource();
+The free function `this_system::resources` is provided for retrieving the `execution_resource`s which encapsualte the hardware platforms available within the system, these are refered to as the *system level resources*.
+
+	std::vector<execution_resource> resources() noexcept;
```
Ruyk: use an alias

affinity/cpp-20/d0796r1.md
```diff
@@ -165,94 +206,163 @@ There are some additional challenges which we have been investigating but are no
...
+
+*Returns:* A std::vector containing all *system level resources*.
+
+*Requires:* If `resources().size() > 0`, `resources()[0]` be the `execution_resource` corroponding to the current thread of execution. The value returned by `resources()` be the same at any point after the invocation of `main`.
```
Ruyk: corroponding -> corresponding
resources.size() must at least be 1, and that must be the resource used for std::thread

affinity/cpp-20/d0796r1.md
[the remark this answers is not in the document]
AerialMantis: I've moved this to the next paragraph which lists the challenges which aren't covered by this paper.

affinity/cpp-20/d0796r1.md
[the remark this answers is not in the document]
AerialMantis: So this is what the paper currently says, sorry my last comment wasn't very clear. By introducing a new device at runtime I meant after `main`, so with the current wording, any runtime is permitted to modify the system topology as seen by `this_system::resources()` as long as it happens before `main` is called. We may want to loosen that requirement to the first call to `this_system::resources()` in the future if this is problematic, but for now I think this is fine.

affinity/cpp-20/d0796r1.md
AerialMantis: I feel creating an alias for this will make the wording more complex as we'd have to specify what kind of interface the type provides. We have a note that we may want to change this type in the future.

affinity/cpp-20/d0796r1.md
AerialMantis: This has to be a reference, because if it were returned by value then if `memory_resource()` were to be called from two different places (potentially different threads) each of the calls would create a copy of the memory_resource, which means they would not be allocating memory in the same resource.

affinity/cpp-20/d0796r1.md
AerialMantis: Fixed.
See earlier comment on reference counting.

affinity/cpp-20/d0796r1.md
AerialMantis: Fixed.

affinity/cpp-20/d0796r1.md
AerialMantis: You're right this should be by value, I've fixed this.

affinity/cpp-20/d0796r1.md
AerialMantis: Is it not important to clarify that the resource is capable of allocating with affinity, otherwise it's just a regular allocator?

affinity/cpp-20/d0796r1.md
AerialMantis: I wasn't sure how we should define this but I've left this deliberately vague for now as P0737r0 defines properties which specify on destruction behavior, however, these are still being discussed so I didn't want to include them just yet.

affinity/cpp-20/d0796r1.md
AerialMantis: That's an interesting point, though I would expect that an implementation would have a string allocated on construction so that `name` would not have to perform any allocation.

affinity/cpp-20/d0796r1.md
AerialMantis: Done.

affinity/cpp-20/d0796r1.md
```diff
@@ -95,395 +96,423 @@ Some systems give additional user control through explicit binding of threads to
...
-      execution_resource &operator=(execution_resource &&) = delete;
-      ~execution_resource();
+Nowadays, there are various APIs and libraries that enable this functionality. One of the most commonly used is the Portable Hardware Locality (hwloc) [9]. Hwloc presents the hardware as a tree, where the root node represents the whole machine and subsequent levels represent different partitions depending on different hardware characteristics. The picture below shows the output of the hwloc visualization tool (lstopo) on a 2-socket Xeon E5300 server. Note that each socket is represented by a package in the graph. Each socket contains its own cache memories, but both share the same NUMA memory region. Note also that different I/O units are visible underneath: Placement of these units with respect to memory and threads can be critical to performance. The ability to place threads and/or allocate memory appropriately on the different components of this system is an important part of the process of application development, especially as hardware architectures get more complex. The documentation of lstopo [22] shows more interesting examples of topologies that appear on today's systems.
```
mhoemmen: "One of the most commonly used is the Portable Hardware Locality (hwloc) library [9]."

affinity/cpp-20/d0796r1.md
```diff
@@ -95,395 +96,423 @@ Some systems give additional user control through explicit binding of threads to
...
 
-      size_t concurrency() const noexcept;
-      size_t partition_size() const noexcept;
+![alt text](hwloc-topology.png "Hwloc topology")
 
-      const execution_resource &partition(size_t i) const noexcept;
-      const execution_resource &member_of() const noexcept;
+Th interface of `thread_execution_resource_t` proposed in the [execution context proposal][p0737r0] proposes a hierarchical approach where there is a root resource and each resource has a number of child resources. However, systems are becoming increasingly non-hierarchical and a traditional tree-based representation of a **system’s resource topology** may not suffice any more [18]. The HSA standard solves this problem by allowing a node in the topology to have multiple parent nodes [19].
```
mhoemmen: "The interface of"

affinity/cpp-20/d0796r1.md
AerialMantis: Fixed.

affinity/cpp-20/d0796r1.md
AerialMantis: Fixed.

AerialMantis:
@mhoemmen thanks for the feedback.

affinity/cpp-20/d0796r1.md
```diff
@@ -1,29 +1,30 @@
...
 
 # Changelog
 
-### P0796R1 (JAX)
+### P0796r1 (JAX)
 
 * Introduce proposed wording.
-* Have `execution_context::resource()` and `this_system::resource()` return a `const execution_resource &`.
+* Based on feedback from SG1, introduced a pair-wise interface for querying the relative affinity between execution resources.
+* Based on feedback from SG1, introduced a non-hierarchical system tpology structure, which doesn't require a root resouce (TODO).
```
Ruyk: why is there a TODO?

Ruyk: also tpology -> topology

affinity/cpp-20/d0796r1.md
```diff
@@ -95,395 +96,423 @@ Some systems give additional user control through explicit binding of threads to
...
 
-    struct execution_resource {
+> For example, a NUMA system will likely have a hierarchy of nodes, each capable of placing memory and placing agents.  A CPU + GPU system may have GPU local memory regions capable of placing memory, but not capable of placing agents.
```
Ruyk: there is >

affinity/cpp-20/d0796r1.md
AerialMantis: The TODO was just a reminder for myself, I've removed it now.

AerialMantis: Fixed the typo.

affinity/cpp-20/d0796r1.md
AerialMantis: This was indentation for the example, but we don't really use this format anywhere else so I'll just remove it.

affinity/cpp-20/d0796r1.md
```diff
@@ -95,395 +96,423 @@ Some systems give additional user control through explicit binding of threads to
...
 
-The `execution_resource` class is not `CopyConstructible` (C++Std [copyconstructible]).
+### Execution resources
 
-    execution_resource(const execution_resource &) = delete;
-    execution_resource(execution_resource &&) = delete;
-    execution_resource &operator=(const execution_resource &) = delete;
-    execution_resource &operator=(execution_resource &&) = delete;
+An `execution_resource` is a light weight structure which acts as an identifier to particuar piece of hardware within a system. It can be queried for whehter it can allocate memory via `can_place_memory` and whether it can execute work via `can_place_agents`, and for it's name via `name`. An `execution_resource` can also represent other `execution_resource`s, these are refered to as being *members of* that `execution_resource` and can be queried via `resources`. Additionally the `execution_resource` which another is a *member of* can be queried vis `member_of`. An `execution_resource` can also be queried for the concurrency it can provide; the totla number of *threads of execution* supported by that *execution_resource* and all resources it represents.
```
Ruyk: particuar -> particular
whehter -> wether
totla -> total

affinity/cpp-20/d0796r1.md
```diff
@@ -95,395 +96,423 @@ Some systems give additional user control through explicit binding of threads to
...
+> [*Note:* Note that an execution resource is not limited to resources which execute work, but also a general resource where no execution can take place but memory can be allocated such as off-chip memory. *--end note*]
 
-The `execution_resource` class is not `Destructible` (C++Std [destructible]).
+> [*Note:* The intention is that the actual implementation details of a resource topology are described in an execution context when required. This allows the execution resource objects to be lightweight objects that serve as identifiers that are only referenced. *--end note*]
 
-	~execution_resource() = delete;
+### System topology
 
-### `execution_resource` operations
+The system topology is made up of a number of system level `execution_resource`s, which can be queried through `this_system::resource` which returns a `std::vector`. The `execution_resources` available within the system can be initialised dynamically by a runtime library, however must be done so before `main` is called, ass after that point the system topology cannot change.
```
Ruyk: ass after -> given that after

affinity/cpp-20/d0796r1.md
```diff
@@ -95,395 +96,423 @@ Some systems give additional user control through explicit binding of threads to
...
+
+```cpp
+for (auto res : execution::this_system::resources()) {
+    std::cout << res.name()  `\n`;
+    std::cout << res.can_place_memory() << `\n`;
+    std::cout << res.can_place_agents() << `\n`;
+    std::cout << res.concurrency() << `\n`;
+}
+```
+*Listing 2: Example of querying all the system level execution resources*
```
Ruyk: aLL -> all ? not sure weird formatting on my screen

affinity/cpp-20/d0796r1.md
```diff
@@ -95,395 +96,423 @@ Some systems give additional user control through explicit binding of threads to
...
 
-*Returns:*
+### Querying relative affinity
 
-      size_t partition_size() const noexcept;
+The `affinity_query` class template provides an abstraction for a relative affinity value between two `execution_resource`s, derived from a particular `affinity_operation` and `affinity_metric`. The `affinity_query` is templated by `affinity_operation` and `affinity_metric` and is constructed from two `execution_resource`s. An `affinity_query` does not mean much on it's own, instead a relative magnitude of affinity can beb queried by using comparison operators. If nessesary the value of an `affinity_query` can also be queried through `native_affinity`, though the return value of this is implementation defined.
```
Ruyk: beb -> be

affinity/cpp-20/d0796r1.md
```diff
@@ -95,395 +96,423 @@ Some systems give additional user control through explicit binding of threads to
...
+
+```cpp
+auto cList = std::execution::this_system::resources();
+// FindASocketResource is a user-defined function that finds a 
+// resource that is a CPU socket in the given resource list
+auto& socket = findASocketResource(cList);
+execution_contexteC{socket} // Associated with the socket
+auto executor = eC.executor(); // By transitivity, associated with the socket too
+auto socketAllocator = eC.allocator(); // Retrieve an allocator to the closest memory node
+std::vector<int, socketAllocator> v1(100);
```
Ruyk:
```
std::vector<int, decltype(socketAllocator)> v1(100, socketAllocator);
```

affinity/cpp-20/d0796r1.md
AerialMantis: Fixed.

affinity/cpp-20/d0796r1.md
AerialMantis: Fixed.

affinity/cpp-20/d0796r1.md
```diff
@@ -95,395 +96,423 @@ Some systems give additional user control through explicit binding of threads to
...
-// resource that is a CPU socket in the given resource list
-auto& socket = findASocketResource(cList);
-execution_contexteC{socket} // Associated with the socket
-auto executor = eC.executor(myFunctor); // By transitivity, associated with the socket too
-auto socketAllocator = eC.allocator(); // Retrieve an allocator to the closest memory node
-std::vector<int, socketAllocator> v1(100);
-std::generate(par.on(executor), std::begin(v1), std::end(v1), std::rand);
-```
-*Listing 8: Example of allocating with affinity to an execution resource*
+The free function `this_system::resources` is provided for retrieving the `execution_resource`s which encapsualte the hardware platforms available within the system, these are refered to as the *system level resources*.
```
Ruyk: encapsualte -> encapsulate

affinity/cpp-20/d0796r1.md
AerialMantis: I think it's okay, it looks weird on my screen as well.

affinity/cpp-20/d0796r1.md
AerialMantis: Fixed.

affinity/cpp-20/d0796r1.md
AerialMantis: Good catch thanks.

affinity/cpp-20/d0796r1.md
AerialMantis: Fixed.

affinity/cpp-20/d0796r1.md
```diff
@@ -503,6 +532,24 @@ With the ability to place memory with affinity comes the ability to define algor
 |------------|
 | Should the interface provide standard algorithms or policies for distributing memory? |
 
+## Level of abstraction
+
+The current proposal provides an interface for querying whether an `execution_resource` can allocate and/or execute work, it can provide the concurrency it supports and it can provide a name. We also provide the `affinity_query` structure for querying the relative affiity metrics between two `execution_resource`s. However this may not be enough information for users to take full advance of the system, they may also want to know what kind of memory is available or the properties by which work is executed. It was decided that attempting to enumerate the various hardware components would not be ideal as that would make it harder for implementers to support new hardware. It has been discussed that a better approach would be to parameterise the additional properties of hardware such that hardware queries could be much more generic.
```
Ruyk: affiity -> afinity

AerialMantis: Fixed.

AerialMantis:
@Ruyk thanks for the feedback.

PR #29, opened 2018-02-07
d0796r1: Fix "Background Research..." section

mhoemmen:
In d0796r1, fix grammar and wording in "Background Research: State of the Art" section.


mhoemmen:
@Ruyk Thanks!

PR #30, opened 2018-02-08
d0796r1: Update "Proposed Wording" section.

mhoemmen:
d0796r1: Update "Proposed Wording" section.
Mostly grammar fixes, though I did update content some in the MPI / PVM bit.
I would like to revisit the content there at some point.

### DIFF
--- a/affinity/cpp-20/d0796r1.md
+++ b/affinity/cpp-20/d0796r1.md
@@ -165,7 +165,7 @@ There are some additional challenges which we have been investigating but are no
 
 ## Class `execution_resource`
 
-The `execution_resource` class provides an abstraction over a system's hardware capable of memory allocation, execution of light weight exeution agents or both.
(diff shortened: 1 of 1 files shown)


mhoemmen:
@Ruyk Thanks!  I'm not sure how much to write about programming models.  It would make sense to include a discussion of dynamic hardware discovery for programming models with broader non-HPC use.  Right now, the discussion feels too HPC-centric.  On the other hand, the discussion isn't central to the paper, so I wouldn't want it to grow to enormous size :-) .  What do you think?

Ruyk:
Yeah I don't think we should put much more effort on this discussion, we have enough background to justify the need at least for some people, and related work that tackles the issue. If we add more background will look like a "state of the art" paper :-) 
I think we should focus on the interface bits for the next C++ mailing so we have something more concrete for people to provide feedback to.

PR #45, opened 2018-04-24
Specialization constant extension for SYCL

Naghasan:
It propose a way to express and use the SPIR-V concept of specialization constant into SYCL.


Ruyk:
Looks good to me

PR #46, opened 2018-04-24
 Extended specialization constant inferface for the SYCL specialization constant extension

Naghasan:
This proposal extend the original specialisation constant interface (https://github.com/codeplaysoftware/standards-proposals/commit/8a62b94a1af03b685e0fcb7af733a7d3d5867bb5) by adding an interface to use the command group handler to generate specialisation constant.


Ruyk:
This looks good to me, is an interesting proposal.

PR #47, opened 2018-04-24
Add Terminology section

hcedwar:


affinity/cpp-20/d0796r2.md
```diff
@@ -170,6 +170,32 @@ This feature could be easily scaled to heterogeneous and distributed systems, as
 
 In this paper we propose an interface for querying and representing the execution resources within a system, queurying the relative affinity metric between those execution resources, and then using those execution resources to allocate memory and execute work with affinity to the underlying hardware. The interface described in this paper builds on the existing initerface for executors and execution contexts defined in the executors proposal [[22]][p0443r4].
 
+### Terminology
+
+An **execution agent** executes work, typically implemented by a *callable*,
```
AerialMantis: I think we should expand on this to also include a definition for thread of execution and how they relate since we use both terms in the paper. We could say something to the effect of a thread of execution is a single flow of execution within the program and an execution agent represents the steps taken within a thread of execution during the lifetime of a (task | piece of work) executing a callable.

affinity/cpp-20/d0796r2.md
```diff
@@ -170,6 +170,32 @@ This feature could be easily scaled to heterogeneous and distributed systems, as
 
 In this paper we propose an interface for querying and representing the execution resources within a system, queurying the relative affinity metric between those execution resources, and then using those execution resources to allocate memory and execute work with affinity to the underlying hardware. The interface described in this paper builds on the existing initerface for executors and execution contexts defined in the executors proposal [[22]][p0443r4].
 
+### Terminology
+
+An **execution agent** executes work, typically implemented by a *callable*, 
+on an **execution resource** of a given **execution architecture**.
```
AerialMantis: We should probably expand on what an execution architecture is, maybe with some examples, I assume this is the device which encapsulates a topology of resources such as a CPU, GPU, etc.

affinity/cpp-20/d0796r2.md
Ruyk: can we not refer this to the existing definitions on the Executors proposal?

AerialMantis:
In https://github.com/codeplaysoftware/standards-proposals/pull/70 we discussed also introducing a term of art for the *execution resource*.

affinity/cpp-20/d0796r2.md
mhoemmen: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0443r7.html 

"An execution context is a program object that represents a specific collection of execution resources and the execution agents that exist within those resources. Execution agents are units of execution, and a 1-to-1 mapping exists between an execution agent and an invocation of a callable function object submitted via the executor."

P0443 is in flux, given P1055 and ongoing movement towards a compromise (see e.g., https://gist.github.com/ericniebler/69a3a632e1d13f7d8f16e0fbd598e42f ).  Thus, it might be best simply to define these terms here, and later revise if needed to point to definitions in whatever compromise or new Executors paper shows up.

Issue #33, opened 2018-04-09
CP013: Provide clarity on the high-level interface


AerialMantis:
There was feedback from Jacksonville that not all users would want to dive into the fine-grained work of querying a system's topology and manually binding resources, and that many users would instead want a higher-level descriptive interface which allows the implementation to decide how to allocate resources based on the user's requirements. There were concerns that we don't yet clearly present what a high-level interface for affinity would look like.

For Rapperswil we should provide clarity on the different levels of interface that we are proposing and their target users, and also on how we envision a high-level interface for affinity C++ to look like, preferably with examples.

Ruyk:
This can be closed after https://github.com/codeplaysoftware/standards-proposals/pull/48 is merged.

PR #48, opened 2018-05-02
Issue #33: Add wording for high-level interface.

AerialMantis:
* Introduce `this_thread::bind` & `this_thread::unbind`.
* Introduce high-level interface `bulk_execution_affinity` properties.

### DIFF
--- a/affinity/cpp-20/d0796r2.md
+++ b/affinity/cpp-20/d0796r2.md
@@ -132,7 +132,7 @@ The interface of `thread_execution_resource_t` proposed in the execution context
 
 The interface for querying the *resource topology* of a *system* must be flexible enough to allow querying all *execution resources* available under an *execution context*, querying the *execution resources* available to the entire system, and constructing an *execution context* for a particular *execution resource*. This is important, as many standards such as OpenCL [[6]][opencl-2-2] and HSA [[7]][hsa] require the ability to query the *resource topology* available in a *system* before constructing an *execution context* for executing work.
 
-> For example, an implementation may provide an execution context for a particular execution resource such as a static thread pool or a GPU context for a particular GPU device, or an implementation may provide a more generic execution context which can be constructed from a number of CPU and GPU devices queryable through the system resource topology.
+> For example, an implementation may provide an execution context for a particular execution resource such as a static thread pool or a GPU context for a particular GPU device, or an implementation may provide a more generic execution context which can be constructed from a number of CPU and GPU devices query-able through the system resource topology.
 
 ### Topology discovery & fault tolerance
 
@@ -188,11 +188,7 @@ The high-level interface is a policy-based design which utilizes the executor pr
 
 ### Bulk execution affinity
 
-In this paper we propose an executor property group called `bulk_execution_affinity` which contains the sub properties `none`, `balanced`, `scatter` or `compact`. Each of these properties, if applied to an *executor* enforce a particular guarantee of execution agent binding to the *execution resources* associated with the *executor* in a particular pattern:
-* **none** makes no guarantee that *execution agents* created by the *executor* will be bound to specific *execution resources*.
-* **balanced** guarantees that *execution agents* created by the executor will be bound to the *execution resources* associated with the *executor* close together in sequence but with an even distribution across the *execution resources*.
-* **scatter** guarantees that *execution agents* created by the executor will be bound to the *execution resources* associated with the *executor* distributed with each *execution agent* far from each other *execution agent* in sequence.
-* **compact** guarantees that *execution agents* created by the executor will be bound to the *execution resources* associated with the *executor* close together in sequence.
+In this paper we propose an executor property group called `bulk_execution_affinity` which contains the nested properties `none`, `balanced`, `scatter` or `compact`. Each of these properties, if applied to an *executor* enforce a particular guarantee of execution agent binding to the *execution resource*s associated with the *executor* in a particular pattern.
 
 Below *(Listing 2)* is an example of executing a parallel task over 8 threads using `bulk_execute`, with the affinity binding `bulk_execution_affinity.scatter`.
 
@@ -210,8 +206,6 @@ Below *(Listing 2)* is an example of executing a parallel task over 8 threads us
 
 ```
 *Listing 2: Example of using the bulk_execution_affinity property*
-
-> [*Note:* The terms used for the `bulk_execution_affinity` property group are derived from the OpenMP properties [[33]][openmp-affinity] including the Intel specific balanced affinity binding [[[34]][intel-balanced-affinity] *--end note*]
 
 ## Low-level interface
 
@@ -319,6 +313,12 @@ A *thread of execution* can be requested to bind to a particular `execution_reso
     namespace experimental {
     namespace execution {
 
+    /* Bulk execution affinity properties */
+
+    struct bulk_execution_affinity_t;
+
+    constexpr bulk_execution_affinity_t bulk_execution_affinity;
+
     /* Execution resource */
 
     class execution_resource {
@@ -417,6 +417,25 @@ A *thread of execution* can be requested to bind to a particular `execution_reso
     }  // std
 
 *Listing 7: Header synopsis*
+
+## Bulk execution affinity properties
+
+The bulk_execution_affinity_t property describes what guarantees executors provide about the binding of *execution agent*s to the underlying *execution resource*s.
+
+bulk_execution_affinity_t provides nested property types and objects as described below.
+
+| Nested Property Type | Nested Property Name | Requirements |
+|----------------------|----------------------|--------------|
+| bulk_execution_affinity_t::none_t | bulk_execution_affinity_t::none | A call to an executor's bulk execution function may or may not bind the the *execution agent*s to the underlying *execution resource*s. The affinity binding pattern may or may not be consistent across invocations of the executor's bulk execution function.  |
+| bulk_execution_affinity_t::scatter_t | bulk_execution_scatter_t::scatter | A call to an executor's bulk execution function must bind the *execution agent*s to the underlying *execution resource*s such that they are distributed across the *execution resource*s where each *execution agent* far from it's preceding and following *execution agent*s. The affinity binding pattern must to be consistent across invocations of the executor's bulk execution function. |
+| bulk_execution_affinity_t::compact_t | bulk_execution_compact_t::compact | A call to an executor's bulk execution function must bind the *execution agent*s to the underlying *execution resource*s such that they are in sequence across the *execution resource*s where each *execution agent* close to it's preceding and following *execution agent*s. The affinity binding pattern must be consistent across invocations of the executor's bulk execution function. |
+| bulk_execution_affinity_t::balanced_t | bulk_execution_balanced_t::balanced | A call to an executor's bulk execution function must bind the *execution agent*s to the underlying *execution resource*s such that they are in sequence and evenly spread across the *execution resource*s where each *execution agent* close to it's preceding and following *execution agent*s and all *execution resource*s are utilized. The affinity binding pattern must to be consistent across invocations of the executor's bulk execution function. |
+
+> [*Note:* The requirements of the `bulk_execution_affinity_t` nested properties do not enforce a specific binding, simply that the binding follows the requirements set out above and that the pattern is consistent across invocations of the bulk execution functions. *--end note*]
+
+> [*Note:* If two *executor*s `e1` and `e2` invoke a bulk execution function in order, where `execution::query(e1, execution::context) == query(e2, execution::context)` is `true` and `execution::query(e1, execution::bulk_execution_affinity) == query(e2, execution::bulk_execution_affinity)` is `false`, this will likely result in `e1` binding *execution agent*s if necessary to achieve the requested affinity pattern and then `e2` rebinding rebinding tp achieve the new affinity pattern. *--end note*]
+
+> [*Note:* The terms used for the `bulk_execution_affinity_t` nested properties are derived from the OpenMP properties [[33]][openmp-affinity] including the Intel specific balanced affinity binding [[[34]][intel-balanced-affinity] *--end note*]
 
 ## Class `execution_resource`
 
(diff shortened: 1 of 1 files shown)


affinity/cpp-20/d0796r2.md
```diff
@@ -170,6 +175,43 @@ This feature could be easily scaled to heterogeneous and distributed systems, as
 
 In this paper we propose an interface for querying and representing the execution resources within a system, queurying the relative affinity metric between those execution resources, and then using those execution resources to allocate memory and execute work with affinity to the underlying hardware. The interface described in this paper builds on the existing initerface for executors and execution contexts defined in the executors proposal [[22]][p0443r4].
 
+### Interface grandularity
+
+In this paper we propose both a low-level interface and a high-level interface:
+* The low-level interface cosnsists of mechanisms for discovering detailed information about a system's topology and affinity properties which can be utilised to hand optimise parallel applications and libraries for the best performance. The low-level interface has high granularity and is aimed at users who have a high knowledge of the system architecture.
```
hcedwar: typo "cosnsists"

affinity/cpp-20/d0796r2.md
```diff
@@ -252,14 +292,20 @@ auto socketAllocator = eC.allocator(); // Retrieve an allocator to the closest m
...
 
 The construction of an `execution_context` on a component implies affinity (where possible) to the given resource. This guarantees that all executors created from that `execution_context` can access the resources and the internal data structures requires to guarantee the placement of the processor.
 
 Only developers that care about resource placement need to care about obtaining executors and allocations from the correct `execution_context` object. Existing code for vectors and STL (including the Parallel STL interface) remains unaffected.
 
 If a particular policy or algorithm requires to access placement information, the resources associated with the passed executor can be retrieved via the link to the `execution_context`.
 
+### Binding to execution
+
+A *thread of execution* can be bound to a particular `execution_resource` for a particular *execution agent* by calling `this_thread::bind`. After which point the *execution resource* returned by `this_thread::get_resource` must be equal to the `execution_resource` provided to `this_thread::bind`. Subsequently a *thread of execution* can be unbound by calling `this_thread::unbind`.
```
hcedwar: mention only if _can bind_

affinity/cpp-20/d0796r2.md
```diff
@@ -515,6 +566,20 @@ The free function `this_system::resources` is provided for retrieving the `execu
 
 > [*Note:* Returning a `std::vector` allows users to potentially manipulate the container of `execution_resource`s after it is returned, we may want to replace this with an alternative type which is more restrictive at a later date such as a range. *--end note*]
 
+The free functions `this_thread::bind` and `this_thread::unbind` are provided for binding / unbinding the current *thread of execution* to / from a particular `execution_reosurce`.
```
hcedwar: typo "reosurce"

affinity/cpp-20/d0796r2.md
```diff
@@ -515,6 +566,20 @@ The free function `this_system::resources` is provided for retrieving the `execu
...
+
+    bool bind(executon_resource) noexcept;
```
Ruyk: executon -> execution

affinity/cpp-20/d0796r2.md
```diff
@@ -515,6 +566,20 @@ The free function `this_system::resources` is provided for retrieving the `execu
...
+
+*Returns:* `true` if the requested binding was successful, otherwise `false`.
+
+*Effects:* If successful, binds the current *thread of execution* to the specified `execution_resource`.
+
+    bool unbind(executon_resource) noexcept;
```
Ruyk: executon -> execution

affinity/cpp-20/d0796r2.md
```diff
@@ -515,6 +566,20 @@ The free function `this_system::resources` is provided for retrieving the `execu
...
+
+*Returns:* `true` if the requested unbinding was successful, otherwise `false`.
```
Ruyk: Is true and false enough information on the error? is there any particular reason for not  to throw an exception?

affinity/cpp-20/d0796r2.md
```diff
@@ -170,6 +175,43 @@ This feature could be easily scaled to heterogeneous and distributed systems, as
...
+
+### Bulk execution affinity
+
+In this paper we propose an executor property group called `bulk_execution_affinity` which contains the sub properties `none`, `balanced`, `scatter` or `compact`. Each of these properties, if applied to an *executor* enforce a particular guarantee of execution agent binding to the *execution resources* associated with the *executor* in a partuclar pattern:
+* **none** makes no guarantee that *execution agents* created by the *executor* will be bound to specific *execution resources*.
+* **balanced** guarantees that *execution agents* created by the executor will be bound to the *execution resources* associated with the *executor* close together in sequence but with an even distribution across the *execution resources*.
+* **scatter** guarantees that *execution agents* created by the executor will be bound to the *execution resources* associated with the *executor* distributed with each *execution agent* far from each other *execution agent* in sequence.
+* **compact** guarantees that *execution agents* created by the executor will be bound to the *execution resources* associated with the *executor* close together in sequence.
```
Ruyk: if this is based on the OpenMP binding, we should add a reference

affinity/cpp-20/d0796r2.md
```diff
@@ -170,6 +175,43 @@ This feature could be easily scaled to heterogeneous and distributed systems, as
...
+{
+  auto exec = executionContext.executor();
+
+  auto affExec = execution::require(exec, execution::bulk,
+    execution::bulk_execution_affinity.scatter);
+
+  affExec.bulk_execute([](std::size_t i, shared s) {
+    func(i);
+  }, 8, sharedFactory);
+}
```
Ruyk: Let me see if I understand this properly. Assuming this is a simple fixed size thread pool of size 4 underneath:

* The thread pool itself is created and "maintained" by the execution context. That means the threads are created on construction of the execution context. At this points, threads are bound to wherever execution(s) resource(s) have been indicated on construction of the executionContext, if any.
* Then we perform a require for a bulk executor following a scatter policy. What do you expect to happen?
A. The threads of the thread pool are re-bound following the scatter placement of threads per resource
B. The new _affExec_ will enforce placement of execution agents on the thread pool threads following the scatter policy.
C. Neither of the above
My understanding is B from the proposal.
However, if what I want to do is place the actual execution threads following scatter policy on the given cores, I would need to pass the policy on construction of the execution context, rather than the executor (since the thread pool has been created already). I could potentially re-bind threads after they have been created, but that has a cost that could be avoided if the initial placement of resources of the thread pool is done on construction. Can we have an alternative constructor where these policies are passed to the execution context? If we are having a fine-grain selection of member_of resources, this will allow for the high-level interface to be used in that case.

* Now we call bulk_execute with the callable. How is each execution agent placed onto an execution resource? If the executor property is _thread_execution_mapping_t_ or other_execution_mapping_t , that means the executor will query the placement (presumably an id) and place agents on existing threads. Which one is a valid placement?
A. Agent 0 in Thread 0, Agent 1 in Thread 2, Agent 2 in Thread 1, Agent 3 in Thread 3, then loop over again for the remaining 4 agents
B. Agent 0 in Thread 0, Agent 1 in Thread 3, Agent 2 in Thread 1, Agent 3 in Thread 3, then loop over again for the remaining 4 agents
C. Either of the above
My understanding is A will be correct, or at least commonly expected. 
However, when oversubscribing agents on threads is not clear what to expect when using these policies. Will the executor hold execution of an agent until the resource "far away" is available or will execute consecutively if possible?
Also note that the same execution context (in this case, the same thread pool) may be in use by multiple executors. What is the expectation in terms of placement of agents on threads?

affinity/cpp-20/d0796r2.md
```diff
@@ -14,6 +14,11 @@
 
 # Changelog
 
+### P0796r2 (RAP)
+
+* Introduce `this_thread::bind` & `this_thread::unbind` for binding a thread of execution to an execution resource.
```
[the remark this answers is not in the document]
AerialMantis: Done.

affinity/cpp-20/d0796r2.md
```diff
@@ -170,6 +175,43 @@ This feature could be easily scaled to heterogeneous and distributed systems, as
 
 In this paper we propose an interface for querying and representing the execution resources within a system, queurying the relative affinity metric between those execution resources, and then using those execution resources to allocate memory and execute work with affinity to the underlying hardware. The interface described in this paper builds on the existing initerface for executors and execution contexts defined in the executors proposal [[22]][p0443r4].
 
+### Interface grandularity
```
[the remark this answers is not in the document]
AerialMantis: Done.

affinity/cpp-20/d0796r2.md
AerialMantis: Fixed.

affinity/cpp-20/d0796r2.md
```diff
@@ -170,6 +175,43 @@ This feature could be easily scaled to heterogeneous and distributed systems, as
...
+* The high-level interface consists of policies which describe desired behaviour when using parallel algorithms or libraries. The high-level interface has low granularity and is aimed at users who may have little or no knowledge of the system architecture.
+
+## High-level interface
+
+The high-level interface is a polcy-based design which utilises the executor property mechanism to provide additional affinity based requirements on executors.
+
+### Bulk execution affinity
+
+In this paper we propose an executor property group called `bulk_execution_affinity` which contains the sub properties `none`, `balanced`, `scatter` or `compact`. Each of these properties, if applied to an *executor* enforce a particular guarantee of execution agent binding to the *execution resources* associated with the *executor* in a partuclar pattern:
```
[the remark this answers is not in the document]
AerialMantis: Fixed.

affinity/cpp-20/d0796r2.md
```diff
@@ -170,6 +175,43 @@ This feature could be easily scaled to heterogeneous and distributed systems, as
...
+* **none** makes no guarantee that *execution agents* created by the *executor* will be bound to specific *execution resources*.
+* **balanced** guarantees that *execution agents* created by the executor will be bound to the *execution resources* associated with the *executor* close together in sequence but with an even distribution across the *execution resources*.
```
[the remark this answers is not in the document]
AerialMantis: I chose `balanced` as that was the term used in OpenMP, though I'm open to alternatives, I expect these to be bikeshed in the future.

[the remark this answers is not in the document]
AerialMantis: I actually think the `balanced` affinity property is specific to Intel's OpenMP, i'll note this in the paper.

affinity/cpp-20/d0796r2.md
AerialMantis: Yes, this is based on OpenMP binding, I will add a note for this with links.

affinity/cpp-20/d0796r2.md
AerialMantis: @hcedwar Thanks I missed that, I've added a requirement that the `execution_resource` be capable of placing agents.
@Ruyk Thanks for pointing this out, I have added some wording which describes what happens when binding or unbinding is unssucessfull. For now, I avoided having it throw an exception just to avoid a barrier to implementation on architectures which don't support exceptions, though this may change in the future. Yes that's right, the idea is that after a thread of execution is successfully bound to an execution resource, any time you query that thread of execution for it's resource you should get the same resource you bound to it.

affinity/cpp-20/d0796r2.md
```diff
@@ -252,14 +292,20 @@ auto socketAllocator = eC.allocator(); // Retrieve an allocator to the closest m
...
 
 Only developers that care about resource placement need to care about obtaining executors and allocations from the correct `execution_context` object. Existing code for vectors and STL (including the Parallel STL interface) remains unaffected.
 
 If a particular policy or algorithm requires to access placement information, the resources associated with the passed executor can be retrieved via the link to the `execution_context`.
 
+### Binding to execution
+
+A *thread of execution* can be bound to a particular `execution_resource` for a particular *execution agent* by calling `this_thread::bind`. After which point the *execution resource* returned by `this_thread::get_resource` must be equal to the `execution_resource` provided to `this_thread::bind`. Subsequently a *thread of execution* can be unbound by calling `this_thread::unbind`.
+
+> [*Note:* Binding *threads of execution* can provide performance benefits when used in a way which compliments the application, however incorrect usage can lead to denial of service and therefore can cause loss of performance. *--end note*]
```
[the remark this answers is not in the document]
AerialMantis: This can be caused if users bind multiple threads of execution which for example, perform busy wait loops.

affinity/cpp-20/d0796r2.md
```diff
@@ -349,11 +395,16 @@ If a particular policy or algorithm requires to access placement information, th
       std::vector<execution_resource> resources() noexcept;
     }
 
+    namespace this_thread {
+        bool bind(executon_resource) noexcept;
```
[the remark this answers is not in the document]
AerialMantis: Yes, that's correct.

affinity/cpp-20/d0796r2.md
AerialMantis: Fixed.

affinity/cpp-20/d0796r2.md
AerialMantis: Fixed.

affinity/cpp-20/d0796r2.md
AerialMantis: Fixed.

affinity/cpp-20/d0796r2.md
AerialMantis: The way I see this working is that if you were to construct an execution context, say the `static_thread_pool` and initialise it with a set number of threads of execution, such as 4. By default, the `bulk_execution_affinity` property would be `none` so the thread pool wouldn't be required to make any kind of guarantee as to how it's threads of execution are bound to the underlying resources. It could automatically bind each of its threads of execution to a certain resource if it makes sense for it to do so, but it wouldn't have to. It would only be when a different `bulk_execution_affinity` property was requested by an executor that it would be required to perform binding in a particular pattern. So when this happens if the thread pool had already performed binding then it may have to rebind to achieve the binding pattern requested by the executor.

The reason for having these properties on the executor rather than the execution context is to make it more usable as a high-level feature. There are some executor implementations which will not make use of an execution context, such as inline executors which are just created as concrete executor types and used without referring to an explicit execution context, an execution context still exists though it's implicit.

I think we should have a more fine-grained interface for configuring execution contexts where you can specify specific affinity binding on construction. In this case, the executor which the execution context provides would only support the `bulk_execution_affinity` property which the execution context was constructed with, so could not be requested to use another, essentially becoming query only. So essentially this would mean that you could specify affinity binding at the execution context level or the executor level, but specifying it at the execution context level would take precidence over the executor level as the executor would inherit the property.

In terms of how the affinity binding is implemented within `bulk_execute`, I would also agree that A would be the most commonly expected, though I think it would be okay to allow an implementation to decide, providing it matched the requirements of the pattern and was consistent across invocations of `bulk_execute`. For oversubscribing agents I would expect that you would have multiple agents mapped to the same threads of execution, so only one agent could progress at a time. As to how those agents are ordered I think that would be at the discretion of the implementation, though implementations would be expected to order agents in way which is most efficient for the requesting binding pattern.

You raise an interesting point here in how an execution context will deal with multiple executors submitting work to it.  In terms of affinity binding patterns, I think I would expect that if one executor was to request one `bulk_execution_affinity` property and another executor was to request a different property then one task would have to be scheduled before the other with the affinity binding pattern being altered between tasks. Though we would recommend users not do this frequently as it could be inneficient to keep rebinding the threads of execution. If both tasks were to be using the same affinity binding pattern, then they could conceivably be overlayed, though I think this would be at the discretion of the implementation as to whether this could be done efficiently.

Ruyk:
```
 So essentially this would mean that you could specify affinity binding at the execution context level or the executor level, but specifying it at the execution context level would take precedence over the executor level as the executor would inherit the property.
```

I guess the opposite would make more sense: The one closer to the user is the one that should make the final decision.

I guess my concern is that associating the binding with the executor may incur in extra costs, since the executor dispatch may be called in a part of the code that is performance critical, whereas the executor context can be defined anywhere.

Apart from that, what you says makes sense to me. Since this is an exploratory paper, I am happy if this is merged with the minor change above and just add a straw poll for deciding where we should do the affinity (executor context, executor or both).

AerialMantis: Yeah, that's a good point, perhaps what we want then is for the executor to be able to override the behaviour of the execution context. Maybe then the property of the execution context could be inherited by the executor as the default behaviour, but can still be altered by the executor.

I think that's a fair concern, that putting control over affinity binding at the executor level could incur costs at the wrong time, though if we also add the ability to configure the execution context as well with the same property then this should alleviate that, as the cost will be paid at configuration time.

Okay great, I will add some notes to the paper covering some of the points we discussed here and I will add a straw poll for where the affinity binding capability should go.

Issue #49, opened 2018-05-02
CP013: Remove polymorphic allocator


AerialMantis:
We decided that the polymorphic allocator does fit the requirements we have for affinity-based allocation, so we should drop the `pmr_memory_resource_type` from the paper and just leave the `allocator_type`.

mhoemmen:
@AerialMantis @hcedwar How should we do this in `execution_context`?  Should the `memory_resource` method still exist, and if so, should it return `allocator_type`?  Or, is the idea that the allocator (instance) behaves as a memory resource?

AerialMantis:
@mhoemmen That's right, originally I was thinking that we remove the `pmr_memory_resource_type` alias and the `memory_resource` function and instead just have the `allocator_type` alias and the `allocator` function, where the allocator instance would behave as the memory resource. Though we've been discussing this again in light of the decision to separate the execution and memory topologies.

**(taken from https://github.com/codeplaysoftware/standards-proposals/pull/52):** *We decided that for the purposes of affinity-based allocation it didn't make sense to use the pmr memory resource, however after separating the execution and memory topologies, I think the pmr memory resource could be a good choice to represent a node within the memory topology, even if that is then passed to a concrete affinity aware allocator type. Though I think we should (as @hcedwar suggested in #41) make it a type which is derived from the pmr memory resource so that it can be made iterable.*

If we were to go down this route perhaps the `memory_resource` function could return the the highest-level memory resource from the memory topology which the execution resources of the execution context can allocate in. In that case you could construct an instance of `allocator_type` with any valid memory resource (i.e. any memory resource under that returned from `memory_resource`) or you could use the allocator instance returned by the `allocator` function which would be initialised with the memory resource returned from `memory_resource`.

What do you think? @mhoemmen @hcedwar @Ruyk

Ruyk:
I think it makes sense. I would like to see something like a memory_context equivalent of the execution_context, but that would not be ready in time for the next meeting.

mhoemmen:
@AerialMantis Sorry I'm a little late with this -- if we separate execution resource and execution context, then I think it makes sense also to separate memory resource and memory context.  This would imply that the default allocator has an implicit memory resource, no?

AerialMantis:
[...]

Issue #41, opened 2018-04-14
Affinity asymmetry between execution and memory resource


hcedwar:
The [Affinity](https://github.com/codeplaysoftware/standards-proposals/blob/master/affinity/cpp-20/d0796r2.md) `execution_context` has asymmetry in that it is constructed from an execution resource but supplies a memory resource.  Should relationship be between execution resources and memory resources without an intervening execution context?

AerialMantis:
We decided to adopt a new approach to representing the system topology where there is a separate topology structure for execution resources and for memory resources.

For this we will need to split out the current `execution_resource` into two classes; for execution resources and for memory resources, and then provide separate ways to query those topologies; for example `this_system::execution_resources` and `this_system::memory_resources`. I think this will also mean dropping `can_place_agents` and `can_place_memory` as it will be implied that execution resources can place agents and memory resources can place memory.

One concern that we may need to consider is that the term memory resource may conflict with the existing pmr memory resource.

AerialMantis:
I was actually thinking that as well that we could used the `pmr::memory_resource` type as our memory resource type, the issue I ran into was having it be traversable as part of a topology. Though I think your solution here of having a topology type which inherits from `pmr::memory_resource` but provides a `begin` and `end` could work quite nicely.

mhoemmen:
Sorry I missed Monday's meeting!  I'm a bit confused about why the memory resource type itself needs to be iterable.  Why wouldn't one just call a function on a memory resource, to get the range of its subordinate resources?  Why would the memory resource itself need begin() and end() instance methods?

mhoemmen:
The reason I ask is that this would, as observed above, require using a subclass of a `pmr` thing, rather than just the `pmr` thing.

mhoemmen:
I get it now -- as long as the resource is a subclass of `pmr`, it can work as a `pmr` thing.  The subclass resource interface is just for iterating hierarchy.

mhoemmen:
However, we did bring up an issue today, that since there's no "memory context," the system could unload a driver when the execution context goes away, and thus break extant memory allocations.

mhoemmen:
Need a lifetime for memory -- e.g., "memory context"

hcedwar:
A "context" for a memory resource is an "allocator"

mhoemmen:
@hcedwar I see -- that makes more sense now.  This would mean users would create an allocator from a memory resource.  Thanks!

[...]

PR #52, opened 2018-05-04
Updates to split execution and memory resources, reduce scope

hcedwar:
Split execution and memory resources within topology.
Make execution resources iterable.
Reduce scope by removing specific execution context, which was not needed for affinity exploration.
Tied current naming and discussion to execution resource capable of executing `std:;thread`.

### DIFF
--- a/affinity/cpp-20/d0796r2.md
+++ b/affinity/cpp-20/d0796r2.md
@@ -12,6 +12,8 @@
 
 **Reply to: gordon@codeplay.com**
 
+**Status: Exploratory**
+
 # Changelog
 
 ### P0796r2 (RAP)
@@ -129,7 +131,7 @@ Nowadays, there are various APIs and libraries that enable this functionality. O
 
 ![alt text](hwloc-topology.png "Hwloc topology")
 
-The interface of `thread_execution_resource_t` proposed in the execution context proposal [[23]][p0737r0] proposes a hierarchical approach where there is a root resource and each resource has a number of child resources. 
+The interface of `thread_execution_resource` proposed in the execution context proposal [[23]][p0737r0] proposes a hierarchical approach where there is a root resource and each resource has a number of child resources. 
 
 Some heterogeneous systems execution and memory resources are not naturally represented by a single tree [[24]][exposing-locality]. The HSA standard solves this problem by allowing a node in the topology to have multiple parent nodes [19].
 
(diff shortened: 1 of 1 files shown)


affinity/cpp-20/d0796r2.md
```diff
@@ -326,9 +327,9 @@ If a particular policy or algorithm requires to access placement information, th
      public:
 
       using native_affinity_type = see-below;
-      using error_type = see-below
+      using error_type = see-below;
 
-      affinity_query(execution_resource &&, execution_resource &&) noexcept;
+      affinity_query(execution_resource & const , std::pmr::memory_resource & const ) noexcept;
```
Ruyk: #49 suggests we should get rid of pmr::memory_resource, should we use simply memory_resource here? Or should we do a separate pull request for that change?

affinity/cpp-20/d0796r2.md
```diff
@@ -474,7 +402,7 @@ The `affinity_query` class template provides an abstraction for a relative affin
 
 ### `affinity_query` constructors
 
-    affinity_query(const execution_resource &, const execution_resource &) noexcept;
+    affinity_query(const execution_resource &, const std::pmr::memory_resource &) noexcept;
```
Ruyk: see my comment above about #49

affinity/cpp-20/d0796r2.md
hcedwar: Let's do a follow up discussion on #49 and follow up PR.

affinity/cpp-20/d0796r2.md
```diff
@@ -110,25 +113,39 @@ There are also some additional challenges which we have been investigating but a
...
+Two important considerations when defining a unified interface for querying the *resource topology* of a *system*, are (a) what level of abstraction such an interface should have, and (b) at what granularity it should describe the topology's *execution resources* and *memory resources*. As both the level of abstraction of resources and the granularity that they describe in will vary greatly from one implementation to another, it’s important for the interface to be generic enough to support any level of abstraction. To achieve this we propose generic hierarchical structures for *execution resources* and *memory resources*, where each *resource* being composed of other *resources* recursively.
```
AerialMantis: I think we hould state here that each execution resource should be made up of other execution resources and each memory resource should be made up of other memory resources:

"where each *resource* being composed of other *resources* recursively" -> "where each is recursively composed of other *execution resources*  and *memory resources* respectively"

affinity/cpp-20/d0796r2.md
```diff
@@ -110,25 +113,39 @@ There are also some additional challenges which we have been investigating but a
...
+The interface of `thread_execution_resource` proposed in the execution context proposal [[23]][p0737r0] proposes a hierarchical approach where there is a root resource and each resource has a number of child resources. 
+
+Some heterogeneous systems execution and memory resources are not naturally represented by a single tree [[24]][exposing-locality]. The HSA standard solves this problem by allowing a node in the topology to have multiple parent nodes [19].
 
-The interface for querying the *resource topology* of a *system* must be flexible enough to allow querying all *execution resources* available under an *execution context*, querying the *execution resources* available to the entire system, and constructing an *execution context* for a particular *execution resource*. This is important, as many standards such as OpenCL [[6]][opencl-2-2] and HSA [[7]][hsa] require the ability to query the *resource topology* available in a *system* before constructing an *execution context* for executing work.
+The interface for querying the *resource topology* of a *system* 
+must be flexible enough to allow
```
AerialMantis: This sentence reads a bit strange, perhaps it's best to split this up now that it's gotten quite long, maybe even have this as bullet points?

affinity/cpp-20/d0796r2.md
```diff
@@ -170,150 +187,134 @@ This feature could be easily scaled to heterogeneous and distributed systems, as
...
 
-### Execution resources
+### Terminology
 
-An `execution_resource` is a light weight structure which acts as an identifier to particular piece of hardware within a system. It can be queried for whether it can allocate memory via `can_place_memory` and whether it can execute work via `can_place_agents`, and for it's name via `name`. An `execution_resource` can also represent other `execution_resource`s, these are refered to as being *members of* that `execution_resource` and can be queried via `resources`. Additionally the `execution_resource` which another is a *member of* can be queried vis `member_of`. An `execution_resource` can also be queried for the concurrency it can provide; the total number of *threads of execution* supported by that *execution_resource* and all resources it represents.
+An **execution agent** executes work, typically implemented by a *callable*,
```
AerialMantis: Is this taken from https://github.com/codeplaysoftware/standards-proposals/pull/47? If so the same comments apply here.

affinity/cpp-20/d0796r2.md
```diff
@@ -170,150 +187,134 @@ This feature could be easily scaled to heterogeneous and distributed systems, as
...
 
-### System topology
 
-The system topology is made up of a number of system level `execution_resource`s, which can be queried through `this_system::resource` which returns a `std::vector`. The `execution_resources` available within the system can be initialised dynamically by a runtime library, however must be done so before `main` is called, given that after that point the system topology cannot change.
+### System topology
 
-Below *(Listing 2)* is an example of iterating over the system level resources and priniting out it's capabilities.
+A *system* includes execution resources and memory resources.
+Execution resources are organized hierarchically.
+A particular execution resource may be a partitioned into
```
AerialMantis: "may be a partitioned into" -> "may be partitioned into"

affinity/cpp-20/d0796r2.md
```diff
@@ -170,150 +187,134 @@ This feature could be easily scaled to heterogeneous and distributed systems, as
...
+a collection of execution resources referred to as *members of*
+that execution resource.
+The partitioning of execution resources implies a locality relationship;
```
AerialMantis: Is it always possible to make this guarantee? Could you, for example, have an architecture which is represented with a hierarchy of {{A, B, C, D},{E, F, G, H}} but where D is more local to E than to A due to the layout of processing units on the die?

affinity/cpp-20/d0796r2.md
```diff
@@ -170,150 +187,134 @@ This feature could be easily scaled to heterogeneous and distributed systems, as
...
+An **execution agent** executes work, typically implemented by a *callable*, 
+on an **execution resource** of a given **execution architecture**. 
+An **execution context** manages a set of execution agents on an 
+execution resource.
+An **executor** submits work to an execution context.
+More that one executor may submit work to an execution context.
+More than on execution context may manage execution agents
+on an execution resource.
 
-> [*Note:* Note that an execution resource is not limited to resources which execute work, but also a general resource where no execution can take place but memory can be allocated such as off-chip memory. *--end note*]
```
AerialMantis: I think this note has been lost in the changes.

affinity/cpp-20/d0796r2.md
Ruyk: Makes sense, lets keep your change until #49 is sorted out.

affinity/cpp-20/d0796r2.md
```diff
@@ -170,150 +187,134 @@ This feature could be easily scaled to heterogeneous and distributed systems, as
...
+  execution::affinity_metric::la1Gtency>(exec1,mem0);
 
-auto relativeLatency = relativeLatency01 > relativeLatency02;
+auto relativeLatency = relativeLatency0 > relativeLatency1;
 ```
 *Listing 3: Example of querying affinity between two `execution_resource`s.*
 
 > [*Note:* This interface for querying relative affinity is a very low-level interface designed to be abstracted by libraries and later affinity policies. *--end note*]
 
-### Execution context
```
AerialMantis: I don't think we should remove the wording for the `execution_context`, this is quite important for the design of being able to construct an execution context which represents part of the system topology.

affinity/cpp-20/d0796r2.md
```diff
@@ -170,150 +187,134 @@ This feature could be easily scaled to heterogeneous and distributed systems, as
...
-If a particular policy or algorithm requires to access placement information, the resources associated with the passed executor can be retrieved via the link to the `execution_context`.
 
 ## Header `<execution>` synopsis
 
     namespace std {
     namespace experimental {
     namespace execution {
 
-    /* Execution resource */
+    /* Execution resource capable of executing std::thread */
```
AerialMantis: I don't think we should couple the execution resource type with `std::thread` as this could restrict from using other kinds of threads of execution.

affinity/cpp-20/d0796r2.md
Ruyk: Seems you are removing entirely the Execution context (I thought it was just moved elsewhere at first). Is that intentional?
The text refers to the concept of Execution context, so I think we should keep it well defined, even if we make changes to the interface itself later.

affinity/cpp-20/d0796r2.md
```diff
@@ -170,150 +187,134 @@ This feature could be easily scaled to heterogeneous and distributed systems, as
...
+      using iterator = /* implementation defined, dereferences to thread_execution_resource */ ;
 
-      execution_resource() = delete;
-      execution_resource(const execution_resource &);
-      execution_resource(execution_resource &&);
-      execution_resource &operator=(const execution_resource &);
-      execution_resource &operator=(execution_resource &&);
-      ~execution_resource();
+      ~thread_execution_resource();
+      thread_execution_resource() = delete;
```
AerialMantis: I think we decided to make the execution resource type copyable and moveable so that it would be easier to use them in standard algorithms.

affinity/cpp-20/d0796r2.md
AerialMantis: I agree. We decided that for the purposes of affinity-based allocation it didn't make sense to use the pmr memory resource, however after separating the execution and memory topologies, I think the pmr memory resource could be a good choice to represent a node within the memory topology, even if that is then passed to a concrete affinity aware allocator type. Though I think we should (as @hcedwar suggested in https://github.com/codeplaysoftware/standards-proposals/issues/41) make it a type which is derived from the pmr memory resource so that it can be made iterable. I think we need to discuss this further.

affinity/cpp-20/d0796r2.md
```diff
@@ -533,23 +461,10 @@ With the ability to place memory with affinity comes the ability to define algor
 |------------|
 | Should the interface provide standard algorithms or policies for distributing memory? |
 
-## Level of abstraction
-
-The current proposal provides an interface for querying whether an `execution_resource` can allocate and/or execute work, it can provide the concurrency it supports and it can provide a name. We also provide the `affinity_query` structure for querying the relative affinity metrics between two `execution_resource`s. However this may not be enough information for users to take full advance of the system, they may also want to know what kind of memory is available or the properties by which work is executed. It was decided that attempting to enumerate the various hardware components would not be ideal as that would make it harder for implementers to support new hardware. It has been discussed that a better approach would be to parameterise the additional properties of hardware such that hardware queries could be much more generic.
+## Dynamic topology discovery - deferred
```
AerialMantis: I think we should leave in the discussion of dynamic topology discovery, as it's an important topic to SG1, we want to show it's still on the roadmap but we just haven't gotten to it yet.

affinity/cpp-20/d0796r2.md
hcedwar: I started down this path and the problem I ran into was having purely hierarchical memory resource.  Even without GPU in the picture you will have main memory, high bandwidth memory, and non-volatile memory.  A hierarchical memory resource would require a top-level allocator that spans all of them.  The path I followed was instead for each execution resource to have a set of memory resources.  
I am uneasy rushing this scope, a little more design-discussion time would have been helpful but we are up against the pre-meeting mailing clock.

affinity/cpp-20/d0796r2.md
hcedwar: I haven't encountered such a thing.  If I did see such a situation I'd say the topology definition was in error.

affinity/cpp-20/d0796r2.md
```diff
@@ -170,150 +187,134 @@ This feature could be easily scaled to heterogeneous and distributed systems, as
...
+A particular execution resource may be a partitioned into
+a collection of execution resources referred to as *members of*
+that execution resource.
+The partitioning of execution resources implies a locality relationship;
+*e.g.*, if `{{A,B},{C,D}}` is a hierarchical partitioning then
+`A` is more local to `B` than it is to `C` or `D`.
+
+An execution resource has one or more memory resources 
+which can allocate memory accessible to that execution resource.
+When an execution resource hierarchy is traversed the associated
```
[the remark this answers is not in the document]
hcedwar: ... prior comment about problem with a pure hierarchy for memory resources

affinity/cpp-20/d0796r2.md
hcedwar: Defines a memory resource vs. an execution resource.

affinity/cpp-20/d0796r2.md
hcedwar: Challenge is what kind of execution context?  Static thread pool?  Dynamic thread pool? etc.

affinity/cpp-20/d0796r2.md
hcedwar: Such a co-routines?
Given the limited time I thought best to start with current-standard foundation of `std::thread` and just associate `thread_execution_resource` .  We can branch out into other architectures later.

affinity/cpp-20/d0796r2.md
hcedwar: The discussion of execution context is still in the paper, just no immediate proposal for a concrete execution context such as static or dynamic thread pools.

affinity/cpp-20/d0796r2.md
hcedwar: I thought this was for the `std::vector<execution_context>` interface, and with iterate-able `thread_execution_resource` was not longer necessary.

affinity/cpp-20/d0796r2.md
hcedwar: Still in the intro material.  Seemed prudent to mention but not spend any SG1 time discussing.

hcedwar:
Rebased pull request the eliminate conflicts.

affinity/cpp-20/d0796r2.md
AerialMantis: That's a good point, having a separate set of memory resources for each execution resource could work, though this could lead to many execution resources pointing to the same memory resource, which could be confusing for users to figure out which memory resources are the same and which ones are different. Though perhaps we could maintain a hierarchy of memory resources, but have each execution resource point to the highest level memory resource which it can allocate memory in, and then the lower level memory resources could be traversed from that it would be consistent enough as to avoid confusion.

One way we could get around the problem of needing to have a top-level allocator could be to have a memory resource which represents the top of the topology which is either not capable of allocating (though that would likely mean throwing an exception) or is free to allocate anywhere within the topology.

I agree I think the separation of execution and memory topologies has revealed some difficult design points which I think we will need some further discussion to resolve.

affinity/cpp-20/d0796r2.md
AerialMantis: That's right, ignore me :)

affinity/cpp-20/d0796r2.md
AerialMantis: I was thinking of GPU threads, though co-routines would be another example of non-`std::thread` threads of execution. As I understand it SG1 are moving away from a tight coupling between `std::thread` and threads of execution, and instead, having `std::thread` just be one of many ways to invoke threads of execution. I feel we should try to follow this direction in our design.

affinity/cpp-20/d0796r2.md
AerialMantis: That's true, but we decided that we should keep resources copyable and moveable so that they could be used in standard algorithms, in order to for example sort by affinity or find particular resources. Although this is something we will need to investigate further. Something like this:

```cpp
std::sort(execRes.begin(), execRes.end(), [&memRes](const foo &lhs, const foo &rhs){
  return affinity_query<read, latency>(lhs, memRes) > affinity_query<read, latency>(rhs, memRes);
});
```

affinity/cpp-20/d0796r2.md
AerialMantis: This is a tough question which I think will require a lot more design discussion. The intention of the current `execution_context` was to provide the minimum required interface in order to allow constructing a context around an execution resource. The idea was that this could then be built on to provide more customization likely through the use of properties.

Perhaps we should move more towards the execution context we define being a concept which requires a particular interface, i.e. one which allows it to be constructed from an execution resource and to provide an executor and allocator. This approach may actually make it easier for concrete execution context types, such as the `static_thread_pool` to support this interface by simply satisfying the requirements of the concept.

Though I'm not keen to make such a large change to the paper so close to the mailing deadline, I think we should raise this as an open question for Rapperswil and then discuss it further for the following revision.

PR #60, opened 2018-05-07
P0796r2: Final changes

AerialMantis:
* Make mionor corrections and fixes.
* Add note to the wording of the `bulk_execution_affinity_t` nestested
properties that requires them to adhere to the requirements of
behavioural properties.
* Rephrase text to remove use of high-level and low-level terms as can
be read ambiguously.
* Add straw poll on the execution context design direction.


fraggamuffin:
LGTM

Issue #50, opened 2018-05-02
CP013: Make resources iterable


AerialMantis:
We decided that in order to avoid having to return a container of member resources from each resource, we should instead make the `execution_resource` (and the subsequent memory resource type) iterable. This means replacing the `resources` member function with `begin` an `end` member functions. Perhaps we should also define iterator traits for the resource types.

AerialMantis:
The MR for this has been merged, so this issue can be closed.

PR #73, opened 2018-08-26
Issue #50: Make execution_resource iterable.

AerialMantis:
* Replace execution_resource::resources with execution_resource::begin
and execution_resource::end.
* Add execution_resource::size.
* Add execution_resource::operator[].
* Add aliases to execution_resource to support new interface.
* Improve woring for execution resource and system topology.
* Update examples to suppoort change.
* Rename this_system::resources to this_system::discover_topology.

### DIFF
--- a/affinity/cpp-20/d0796r3.md
+++ b/affinity/cpp-20/d0796r3.md
@@ -16,6 +16,10 @@
 
 ### P0796r3 (SAN 2018)
 
+* Make `execution_resource`s iterable by replacing `execution_resource::resources` with `execution_resource::begin` and `execution_resource::end`.
+* Add `size` and `operator[]` for `execution_resource`.
+* Rename `this_system::get_resources` to `this_system::discover_topology`.
+
 ### P0796r2 (RAP 2018)
 
 * Introduce a free function for retrieving the execution resource underlying the current thread of execution.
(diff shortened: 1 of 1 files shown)


affinity/cpp-20/d0796r3.md
```diff
@@ -211,26 +215,39 @@ Below *(Listing 2)* is an example of executing a parallel task over 8 threads us
...
 
-> [*Note:* Note that an execution resource is not limited to resources which execute work, but also a general resource where no execution can take place but memory can be allocated such as off-chip memory. *--end note*]
+The **system topology** can be discovered by calling `this_system::discover_topology`. This will discover all **execution resources** available within the system and construct the **system topology** DAG, describing a read-only snapshot at the point of the call, and then return an `execution_resource` object exposing the **system execution resource**.
 
-> [*Note:* The intention is that the actual implementation details of a resource topology are described in an execution context when required. This allows the execution resource objects to be lightweight objects that serve as identifiers that are only referenced. *--end note*]
+> [*Note:* A call to `this_system::discover_topology` may invoke C++ library calls, system calls or third party library APIs required to discover certain **execution resources**. *--end note*]
```
Ruyk: Seems then that discover_topology can throw then?

affinity/cpp-20/d0796r3.md
```diff
@@ -211,26 +215,39 @@ Below *(Listing 2)* is an example of executing a parallel task over 8 threads us
...
 
-The system topology is made up of a number of system-level `execution_resource`s, which can be queried through `this_system::get_resources` which returns a `std::vector`. A run-time library may initialize the `execution_resource`s available within the system dynamically. However, this must be done before `main` is called, given that after that point, the system topology may not change.
+An `execution_resource` object can be queried for a pointer to it's parent `execution_resource` via `member_of`, and can also be iterated over for it's child `execution_resource`s via `begin` and `end`.
 
-Below *(Listing 3)* is an example of iterating over the system-level resources and printing out their capabilities.
+An `execution_resource` object can also be queried for the amount concurrency it can provide, the total number of **threads of execution** supported by the associated **execution resource**.
+
+> [*Note:* An **execution resource** is not limited to resources which execute work, but also a general resource where no execution can take place but memory can be allocated such as off-chip memory. *--end note*]
```
Ruyk: allocated such as -> allocated, such as

affinity/cpp-20/d0796r3.md
```diff
@@ -264,27 +280,27 @@ The `execution_context` class provides an abstraction for managing a number of l
...
 
-auto &systemLevelResource = execContext.resource();
+auto &execResource = execContext.resource();
 
-// resource[0] should be equal to execResource
+// systemResource[0] should be equal to execResource
 
-for (auto res : systemLevelResource.resources()) {
-    std::cout << res.name()  << `\n`;
+for (const execution::execution_resource res : execResource()) {
```
Ruyk: extra () at the end of execResource,
also why const but no &?

affinity/cpp-20/d0796r3.md
```diff
@@ -468,25 +497,37 @@ The `execution_resource` class provides an abstraction over a system's hardware,
...
+    const_iterator begin() const noexcept;
 
-*Returns:* All `execution_resource`s which are *members of* this resource.
+*Returns:* A const iterator to the beggining of the child `execution_resource`s.
+
+    const_iterator end() const noexcept;
+
+*Returns:* A const iterator to the end of the child `execution_resource`s.
+
+    const_reference operator[](int child) const noexcept;
```
Ruyk: shouldn't this be size_type child instead of int?

affinity/cpp-20/d0796r3.md
```diff
@@ -468,25 +497,37 @@ The `execution_resource` class provides an abstraction over a system's hardware,
...
+
+*Returns:* A const reference to the specified child `execution_resource`s.
```
Ruyk: Note this requires random access to the underlying container.

affinity/cpp-20/d0796r3.md
```diff
@@ -589,17 +630,19 @@ The `affinity_query` class template provides an abstraction for a relative affin
...
+The free function `this_system::discover_topology` is provided for discovering the **system topology**.
 
-The free function `this_system::get_resources` is provided for retrieving the `execution_resource`s which encapsulate the hardware platforms available within the system. We refer to these resources as the *system level resources*.
+	const execution_resource discover_topology();
 
-	std::vector<execution_resource> resources() noexcept;
+*Returns:* An `execution_resource` object exposing the **system execution resource**.
 
-*Returns:* An `std::vector` containing all *system level resources*.
+*Requires:* If `this_system::discover_topology().size() > 0`, `this_system::discover_topology()[0]` be the `execution_resource` use by `std::thread`. Calls to `this_system::discover_topology()` may not introduce a data race with any other call to `this_system::discover_topology()`.
```
Ruyk: Does that mean the implementation is responsible to implementing mutex or similar to ensure no data races?
This means the call to discover topology is (even more) expensive. Seems to me, since this is a function that will be called by the library programmer or the programmer on controlled situations, its better for them to handle the control of accessing this function.

affinity/cpp-20/d0796r3.md
```diff
@@ -589,17 +630,19 @@ The `affinity_query` class template provides an abstraction for a relative affin
...
 
-*Requires:* If `this_system::get_resources().size() > 0`, `this_system::get_resources()[0]` be the `execution_resource` use by `std::thread`. The value returned by `this_system::get_resources()` be the same at any point after the invocation of `main`.
+*Effects:* Discovers all **execution resources** available within the system and constructs the **system topology** DAG, describing a read-only snapshot at the point of the call.
 
-> [*Note:* Returning a `std::vector` allows users to potentially manipulate the container of `execution_resource`s after it is returned. We may want to replace this at a later date with an alternative type which is more restrictive, such as a range or span. *--end note*]
+*Throws:* Any exception thrown as a result of **system topology** discovery.
```
Ruyk: So if this throws, does that mean the system topology is unusable? what can the user do to handle that? Either terminate or assume a sequential execution. Probably worth adding some questions for the committee as to what to do when the discovery fails, and what users would expect. Since the discover topology can call different libraries at different points in time, it may be that the first time is called works, but the second doesnt (if there is another _initialize_library_ in between). Even more interesting is the opposite, a call that fails, then a user disables the library, and then discover_topology works again. This should be described or clarified at some point (again, possibly not in this proposal!)

affinity/cpp-20/d0796r3.md
AerialMantis: Yes, in the current proposal `discover_topology` can throw in the case of a failure discovering the topology. Tbh I have concerns about this myself, particularly in the case where the topology discovery is query third-party libraries such as Hwloc or OpenCL. We had an earlier discussion about the idea of partial errors, where calling `discover_topology` could be permitted to fail (either via an exception or some error value tbd) but still return a valid topology structure.

I've added a note to the current paper and added a issue to discuss this further (see https://github.com/codeplaysoftware/standards-proposals/issues/79).

affinity/cpp-20/d0796r3.md
AerialMantis: Fixed.

affinity/cpp-20/d0796r3.md
AerialMantis: Good catch, thanks.

affinity/cpp-20/d0796r3.md
AerialMantis: This is a good point, I've added a requirement to the wording for `iterator` and `const_iterator` to be random access iterators.

affinity/cpp-20/d0796r3.md
AerialMantis: Fixed.

affinity/cpp-20/d0796r3.md
AerialMantis: Yes, the current wording requires that `discover_topology` remain race free with respect to other calls to `discover_topology`. This prevents two simultaneous calls to `discover_topology` from interfering with each other and potentially trying to initialise or finalise the same third-party APIs at the same time. We felt this a fairly low burden to users as topology discovery will generally happen away from the critical path and this will only incur a cost if users were to have multiple threads potentially capable of performing topology discovery. And it means users can safely do this if they want.

This doesn't cover the case where users may manually call a third-party API which invokes the same APIs as `discover_topology`, however, we felt this is not something we can protect users from, and implementations should accurately document any dependencies they have.

This also doesn't currently cover is any potential races between using a resource being used (for example in some execution context) whilst performing topology discovery. This is a slightly trickier case to resolve which we will have to look at.

affinity/cpp-20/d0796r3.md
AerialMantis: Yes at the moment this is the case. This is something which we've discussed before, but I think we should spend some more time discussing. I responded on this in more detail in a comment above and created an issue for discussing it further (see https://github.com/codeplaysoftware/standards-proposals/issues/79).

Issue #67, opened 2018-07-24
CP013: Calrify the requirements for the copy and move semantics of the `execution_resource`


AerialMantis:
Currently, the requirements for the `execution_resource` are quite vague:

> [Note: The intention is that the actual implementation details of a resource topology are
described in an execution context when required. This allows the execution resource objects
to be lightweight objects that serve as identifiers that are only referenced. --end note]

In #40 we decided that the `execution_resource` should be copyable and moveable but must act as an opaque type with reference counting semantics.

Taken from #40:
> * Answering the second point, the `execution_reosurce` should remain copyable and moveable so that it can be used within std alogrithms, but it should be an opaque type with reference counting semantics.

Perhaps we want to introduce normative wording which requires certain behaviour of the `execution_resource` when being copied or moved in order to guarantee the corerect behaviour.

mhoemmen:
I wish I had been there for the Monday telecon.  David Hollman and I discussed this on Friday and we came up with the following concerns:

  1. We couldn't think of any type in the Standard that has reference-counting semantics, but doesn't have `shared_` in the name (e.g., `shared_ptr`, `shared_future`).
  2. Reference counting could have unattractive overhead if accessed concurrently.
  3. What about architectures with execution resources that might only exist during a parallel execution?
  4. It's not clear that all kinds of execution resources would fit naturally into a reference-counting lifetime model.  For example, what if an operating system can "take back" available execution resources during a program's execution, as long as the program isn't currently using them?  Any view of the execution resources might be just a snapshot of what's currently available.

I was thinking that it's really just enough for execution resources to be unique identifiers.  That is, if I view a subset of available resources, and then request to create a context with that subset, those identifiers shouldn't change between my first view and the context creation request.

David had other ideas too -- I'm not speaking for him here, I just wanted to get some ideas out there before the Monday morning meeting.

mhoemmen:
Notes from Monday 30 July 2018 telecon:

The point of making execution resources copyable and moveable is so that we can use standard algorithms to traverse the hierarchy of resources, e.g., for depth-based views of topology.  This also enables use of the "pImpl" idiom to hide the implementation.  @mhoemmen agrees that this makes sense.

[...]

PR #74, opened 2018-09-10
Affinity: Address #67 & discussion in #70

mhoemmen:
@AerialMantis 

Address #67 by applying discussion in #70.
Update some references, including p0443.


affinity/cpp-20/d0796r3.md
```diff
@@ -221,7 +276,7 @@ An `execution_resource` is a lightweight structure which acts as an identifier t
 
 ### System topology
 
-The system topology is made up of a number of system-level `execution_resource`s, which can be queried through `this_system::get_resources` which returns a `std::vector`. A run-time library may initialize the `execution_resource`s available within the system dynamically. However, this must be done before `main` is called, given that after that point, the system topology may not change.
+The system topology is made up of a number of system-level `execution_resource`s, which can be queried through `this_system::get_resources` which returns a `std::vector`. A run-time library may initialize the `execution_resource`s available within the system dynamically. However, `this_system::get_resources` must be thread safe and must initialize and finalize any third-party or OS state before returning.
```
Ruyk: If we don't want iterators to be invalidated, shouldn't this be std::list ?

affinity/cpp-20/d0796r3.md
```diff
@@ -629,16 +684,19 @@ The free function `this_thread::get_resource` is provided for retrieving the `ex
 
 *Returns:* The `execution_resource` underlying the current thread of execution.
 
+> [*Note:* The `execution_resource` underlying the current thread of execution may not necessarily be reachable from "top-level" resources visible through `this_system`. *--end note*]
```
Ruyk: Is not clear to me what is the reason for the note (i have been on holiday so apologies if this has been discussed already)

affinity/cpp-20/d0796r3.md
mhoemmen: Issue #50 and PR #73 address making resources iterable.  I wanted this PR to be as orthogonal from those as possible, though I think I did manage to conflict with #73, alas ;-) .  Anyway, thanks for pointing this out!

affinity/cpp-20/d0796r3.md
mhoemmen: One of the things we discussed was whether some resources might not be accessible from the top level -- e.g., some GPU resources (like scratch memory) might only be available during execution on the GPU.  One could thus access them through `this_thread::get_resource` (and possibly in other ways), but not from the top level `this_system::get_resource`.

This is also one of the reasons for `execution_resource` not to be reference counted (at least not all `execution_resource`s).  Otherwise, it would be bad for them to escape the scope of code executing on the GPU.

I was thinking about this Note last night, and realized that we may need more explanatory text about this.  I'd like the PR to go through so it doesn't step on other PRs, but we should definitely write more about this.

AerialMantis:
This looks great, thanks!

PR #75, opened 2018-09-14
CP016: Sub groups proposal

AerialMantis:
Add a proposal for introducing an extension to support sub-groups in SYCL 1.2.1.


sub-groups/sycl-1.2.1/index.md
```diff
@@ -0,0 +1,157 @@
...
+
+Sub-groups can access global and local memory, but, given there is no 
+memory-scope to the atomic or barriers operations in SYCL 1.2.1, there is no
+possibility to specify an equivalent of sub-group memory scope.
+
+## Namespace `basic_sub_group`
+
+All new functionality is exposed under the `basic_sub_group` namespace
+in the `codeplay` vendor extension namespace.
+When the vendor extension `basic_sub_group` is available, the macro
```
alycm: OpenCL subgroup extensions used `subgroups` in their names (e.g. `cl_khr_subgroups`).

It may be valid to ignore that given the type `cl::sycl::codeplay::sub_group`.

sub-groups/sycl-1.2.1/index.md
```diff
@@ -0,0 +1,157 @@
...
+## Execution model
+
+When this vendor extension is available, the execution model of SYCL 1.2.1
+is extended to also include support for sub-groups of threads inside of a
+work-group.
+Overall, these sub-groups work following the description of the OpenCL 2.2
+sub-groups, with some restrictions:
+
+* The number of sub-groups available for each work-group is determined 
+at compile-time and remains the same during the execution of the SYCL application.
```
alycm: I'm not as familiar with the details of the SYCL spec, but looking at it now it seems that current use of "compile-time" means when compiling the SYCL program rather than the SYCL runtime calling `clCompileProgram` or similar.  This information can't be known until `clGetKernelSubGroupInfo` (or `clGetKernelSubGroupInfoCODEPLAY`) can be called.

I'm just unclear on the intended meaning of `compile-time`.

sub-groups/sycl-1.2.1/index.md
```diff
@@ -0,0 +1,157 @@
...
+* The number of threads per sub-group is known at compile-time, and remains the
```
alycm: This use of "threads" seems unusual.

"The sub-group size is known ...`

perhaps?

sub-groups/sycl-1.2.1/index.md
```diff
@@ -0,0 +1,157 @@
...
+The extension adds a new class template `sub_group` that identifies the 
+sub group range and the current sub group id. 
+It also for providing sub group barriers.
+
+```cpp
+namespace cl {
+namespace sycl {
+namespace codeplay {
+
+template <int Dimensions>
```
alycm: I see that this comes from the SYCL 2.2 provisional specification, but for my own sanity: `Dimensions` is here as hypothetical future proofing and barring some future vendor extension it will always have the value `1`?

alycm: Or is it important that, for example, with a 3D ND-Range that sub-group also has `Dimension` of `3`, even if the actual sizes are always of the form {N, 1, 1}?

sub-groups/index.md
```diff
@@ -0,0 +1,36 @@
+# Basic sub-group extension
```
alycm: General comment: Seems like this proposal is missing device info properties?

`info::device::max_num_sub_groups` and `info::device::sub_group_independent_forward_progress`

sub-groups/sycl-1.2.1/index.md
```diff
@@ -0,0 +1,157 @@
...
+
+## Memory model
+
+Sub-groups can access global and local memory, but, given there is no 
+memory-scope to the atomic or barriers operations in SYCL 1.2.1, there is no
+possibility to specify an equivalent of sub-group memory scope.
+
+## Namespace `basic_sub_group`
+
+All new functionality is exposed under the `basic_sub_group` namespace
```
alycm: There doesn't seem to be any `basic_sub_group` namespace used in this extension?

sub-groups/sycl-1.2.1/index.md
Naghasan: `compile-time` is this unclear: SYCL compile time or OpenCL.
The text after seems to imply SYCL because it sub-group range is constexpr,  but this could be made more explicit here.

sub-groups/sycl-1.2.1/index.md
```diff
@@ -0,0 +1,157 @@
...
+```cpp
+namespace cl {
+namespace sycl {
+namespace codeplay {
+
+template <int Dimensions>
+class sub_group {
+ public:
+
+  constexpr range<Dimensions> get_sub_group_range() const;
```
Naghasan: I find this really restrictive, you basically ban compilation for generic targets. But I also understand the need if you know your underlying target. Could we have a in-between solution ? i.e. constexpr iff you know the underlying target and it properties

PR #76, opened 2018-09-17
Initial commit of no-host-access extension

Ruyk:
Vendor extension, and potential future SYCL proposal, to enable optimized creation of SYCL buffers

### DIFF
--- a/host_access/sycl-1.2.1/host_access.md
+++ b/host_access/sycl-1.2.1/host_access.md
@@ -16,7 +16,8 @@ This is possible in OpenCL 1.2 by using the CL_MEM_HOST_NO_ACCESS, CL_MEM_HOST_R
 
 A SYCL implementation for OpenCL needs to analyze the access pattern at compile time, or replace the underlying cl_mem object, in order to infer the host access pattern and create buffers with the appropriate flags.
 
-This proposal enables developers to specify the intended use pattern of a buffer, facilitating a SYCL runtime the usage of optimized buffer flags creation.
+This proposal enables developers to specify the intended use pattern of a buffer, allowing
+SYCL implementations of taking advantage of knowing if a given buffer is usable by the host or not.
 
 ## Revisions
 
--- a/host_access/sycl-1.2.1/host_access.md
+++ b/host_access/sycl-1.2.1/host_access.md
@@ -33,10 +33,10 @@ said context.
 namespace codeplay {
 
 enum class host_access_mode {
-  read;
-  read_write;
-  write;
-  none;
+  read,
+  read_write,
+  write,
+  none
 };
 
 struct host_access {
--- a/host_access/sycl-1.2.1/host_access.md
+++ b/host_access/sycl-1.2.1/host_access.md
@@ -34,10 +34,10 @@ said context.
 namespace codeplay {
 
 enum class host_access_mode {
+  none,
   read,
   read_write,
-  write,
-  none
+  write
 };
 
 struct host_access {
@@ -53,7 +53,9 @@ struct host_access {
 ## Example
 
 The following example is a simplification of a reduction kernel 
-that uses a temporary storage to avoid modifying the input buffer.
+that uses temporary storage to avoid modifying the input buffer, and
+then proceeds to manipulate the data on device only.
+
 
 ```cpp
 context deviceContext;
@@ -121,9 +123,11 @@ auto f = [&](handler &h) mutable {
 
 ## Implementation notes
 
-An implementation may decide to not take advantage of this hint and 
+* An implementation may decide to not take advantage of this hint and 
 simply construct a normal buffer object underneath.
 However, if the buffer is accessed on the host, this must raise an error.
+
+* Using a buffer with host access restrictions with the host-device results in undefined behavior.
 
 ## References
 
--- a/host_access/sycl-1.2.1/host_access.md
+++ b/host_access/sycl-1.2.1/host_access.md
@@ -59,7 +59,7 @@ then proceeds to manipulate the data on device only.
 
 ```cpp
 context deviceContext;
-queue q{context};
+queue q{deviceContext};
 range<3> myRange{3, 3, 3};
 /* A normal buffer that holds the data for 
  * the execution.
@@ -81,8 +81,7 @@ std::for_each(sycl_named_policy<example>(otherQueue),
  * performed.
  */
 buffer<float, 1> tmp{myRange, 
-                     property::buffer::context_bound(deviceContext)
-                    | codeplay::property::buffer::host_access(none) };
+                    {property::buffer::context_bound(deviceContext), codeplay::property::buffer::host_access(host_access_mode::none)} };
 
 bool firstIter = true;
 
@@ -95,9 +94,9 @@ do {
     accessor<float, 1, access::mode::read_write, access::target::local>
           scratch(range<1>(local), h);
     
-    if (first) {
+    if (firstIter) {
       h.parallel_for(r, ReductionKernel(aIn, aInOut));
-      first = false;
+      firstIter = false;
     } else {
       h.parallel_for(r, ReductionKernel(aInOut, aInOut));
     }
--- a/host_access/sycl-1.2.1/host_access.md
+++ b/host_access/sycl-1.2.1/host_access.md
@@ -1,0 +1,130 @@
+# Temporary buffer
+
+| Proposal ID | CP017 |
+|-------------|--------|
+| Name | Host access |
+| Date of Creation | 17 September 2017 |
+| Target | SYCL 1.2.1 vendor extension |
+| Current Status | _Work In Progress_ |
+| Reply-to | Ruyman Reyes <ruyman@codeplay.com> |
+| Original author | Ruyman Reyes <ruyman@codeplay.com> |
+
+## Overview
+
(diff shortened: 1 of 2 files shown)


host_access/sycl-1.2.1/host_access.md
```diff
@@ -0,0 +1,130 @@
...
+| Original author | Ruyman Reyes <ruyman@codeplay.com> |
+
+## Overview
+
+Developers in some situations want to create memory that can only be accessed from the device. This can enable the underlying device driver to allocate memory on places where it is more efficient for the device to access than the host. 
+This is possible in OpenCL 1.2 by using the CL_MEM_HOST_NO_ACCESS, CL_MEM_HOST_READ_ONLY or CL_MEM_HOST_WRITE_ONLY flags during buffer creation.
+
+A SYCL implementation for OpenCL needs to analyze the access pattern at compile time, or replace the underlying cl_mem object, in order to infer the host access pattern and create buffers with the appropriate flags.
+
+This proposal enables developers to specify the intended use pattern of a buffer, facilitating a SYCL runtime the usage of optimized buffer flags creation.
```
ProGTX: `facilitating a SYCL runtime the usage of optimized buffer flags creation` -> this statement is a bit strange, maybe `allowing the SYCL runtime to pass the appropriate flags to the created buffer for optimal performance`?

host_access/sycl-1.2.1/host_access.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Interface changes
+
+This proposal adds a new property for buffers, *buffer::property::host_access*.
+A SYCL buffer with the *host_access* property can only be used within one context,
+and its values are only accessible from queues that are associated with
+said context.
+
+```cpp
+namespace codeplay {
```
ProGTX: I presume the `cl::sycl` namespace is implied?

host_access/sycl-1.2.1/host_access.md
```diff
@@ -0,0 +1,130 @@
...
+This proposal enables developers to specify the intended use pattern of a buffer, facilitating a SYCL runtime the usage of optimized buffer flags creation.
+
+## Revisions
+
+This is the first revision of this proposal.
+
+## Interface changes
+
+This proposal adds a new property for buffers, *buffer::property::host_access*.
+A SYCL buffer with the *host_access* property can only be used within one context,
```
ProGTX: Does that mean it is only valid to use it along with the `context_bound` property?

host_access/sycl-1.2.1/host_access.md
```diff
@@ -0,0 +1,130 @@
...
+said context.
+
+```cpp
+namespace codeplay {
+
+enum class host_access_mode {
+  read;
+  read_write;
+  write;
+  none;
```
ProGTX: `none` should probably be first

host_access/sycl-1.2.1/host_access.md
```diff
@@ -0,0 +1,130 @@
...
+This proposal adds a new property for buffers, *buffer::property::host_access*.
+A SYCL buffer with the *host_access* property can only be used within one context,
+and its values are only accessible from queues that are associated with
+said context.
+
+```cpp
+namespace codeplay {
+
+enum class host_access_mode {
+  read;
```
ProGTX: The values should be separated by commas

host_access/sycl-1.2.1/host_access.md
```diff
@@ -0,0 +1,130 @@
...
+  read_write;
+  write;
+  none;
+};
+
+struct host_access {
```
ProGTX: This should be in the `buffer::property` namespace

host_access/sycl-1.2.1/host_access.md
```diff
@@ -0,0 +1,130 @@
...
+};
+
+}  // namespace codeplay
+```
+
+
+## Example
+
+The following example is a simplification of a reduction kernel 
+that uses a temporary storage to avoid modifying the input buffer.
```
ProGTX: `a temporary storage` -> `temporary storage`

host_access/sycl-1.2.1/host_access.md
```diff
@@ -0,0 +1,130 @@
...
+          elem = 1.0;
+  });
+
+/* This temporary memory only needs to exist 
+ * on the context where the operation will be
+ * performed.
+ */
+buffer<float, 1> tmp{myRange, 
+                     property::buffer::context_bound(deviceContext)
+                    | codeplay::property::buffer::host_access(none) };
```
ProGTX: Properties currently require a `property_list`

host_access/sycl-1.2.1/host_access.md
```diff
@@ -0,0 +1,130 @@
...
+auto length = tmp.get_size();
+
+do {
+  auto f = [&](handler &h) mutable {
+    auto aIn = myBuffer.template get_access<access::mode::read>(h);
+    auto aInOut = tmp.template get_access<access::mode::read_write>(h);
+    accessor<float, 1, access::mode::read_write, access::target::local>
+          scratch(range<1>(local), h);
+    
+    if (first) {
```
ProGTX: `first` -> `firstIter `

host_access/sycl-1.2.1/host_access.md
ProGTX: `none`  should be at least `host_access_mode::none`

host_access/sycl-1.2.1/host_access.md
```diff
@@ -0,0 +1,130 @@
...
+
+
+## Example
+
+The following example is a simplification of a reduction kernel 
+that uses a temporary storage to avoid modifying the input buffer.
+
+```cpp
+context deviceContext;
+queue q{context};
```
ProGTX: Should be constructed from `deviceContext`

host_access/sycl-1.2.1/host_access.md
AerialMantis: `property::buffer::context_bound(deviceContext) | codeplay::property::buffer::host_access(none)` -> `{property::buffer::context_bound(deviceContext), codeplay::property::buffer::host_access(host_access_mode::none)}`

host_access/sycl-1.2.1/host_access.md
```diff
@@ -0,0 +1,130 @@
...
+
+This proposal adds a new property for buffers, *buffer::property::host_access*.
+A SYCL buffer with the *host_access* property can only be used within one context,
+and its values are only accessible from queues that are associated with
+said context.
+
+```cpp
+namespace codeplay {
+
+enum class host_access_mode {
```
AerialMantis: Perhaps we should follow the convention of having access based enums in the access namespace, so something like `access::host_access`?

host_access/sycl-1.2.1/host_access.md
```diff
@@ -0,0 +1,130 @@
...
+
+// Data is now in buffer out, can be copied to the host as usual
+
+```
+
+## Implementation notes
+
+An implementation may decide to not take advantage of this hint and 
+simply construct a normal buffer object underneath.
+However, if the buffer is accessed on the host, this must raise an error.
```
AerialMantis: Should this be: `However, if a buffer that has been constructed with the host_access(host_access::node) property is accessed on the host, this must raise an error.`

host_access/sycl-1.2.1/host_access.md
AerialMantis: Will there also be a similar image property?

host_access/sycl-1.2.1/host_access.md
AerialMantis: What happens if there is no `context_bound` property, should this be an error?

host_access/sycl-1.2.1/host_access.md
```diff
@@ -0,0 +1,130 @@
...
+## Revisions
+
+This is the first revision of this proposal.
+
+## Interface changes
+
+This proposal adds a new property for buffers, *buffer::property::host_access*.
+A SYCL buffer with the *host_access* property can only be used within one context,
+and its values are only accessible from queues that are associated with
+said context.
```
AerialMantis: What would happen if the bound context is a host mode context, would it be possible to create a buffer that's not accessible from anywhere?

host_access/sycl-1.2.1/host_access.md
Ruyk: yes

host_access/sycl-1.2.1/host_access.md
Ruyk: No, the two flags are orthogonal, but the example makes use of both to increase performance.

host_access/sycl-1.2.1/host_access.md
Ruyk: We don't propose this extension for images, only for buffers. It could be implemented, the underlying functionality is there.

host_access/sycl-1.2.1/host_access.md
Ruyk: fixed.

host_access/sycl-1.2.1/host_access.md
Ruyk: Mental breakdown with OpenCL and SYCL interfaces! thanks for pointing it out.

host_access/sycl-1.2.1/host_access.md
Ruyk: it doesn't make sense as a flag to an accessor, only makes sense for a buffer (or image) constructor.

Issue #42, opened 2018-04-16
Affinity between execution and memory resource


hcedwar:
[Affinity](https://github.com/codeplaysoftware/standards-proposals/blob/master/affinity/cpp-20/d0796r2.md) has `affinity_query` between two execution resources.  Recommend this be between an execution resource and a memory resource.

hcedwar:
Querying an affinity metric should be between an execution resource and a memory resource.
Querying an affinity metric between two execution resources would be transitive: "what is our affinity with the most-local memory resource we have in common."

pblinzer:
The affinity metric should have also some qualifier attribute mechanisms, e.g. for some heterogeneous execution resources, the availability of global atomicsOps or cc property on "nearest common memory resource" may not be there.

Issue #68, opened 2018-07-30
Expose the magnitude of relative affinity queries through a different interface


AerialMantis:
It was suggested at the Rapperswil meeting that we should consider alternatives to having the `affinity_query` comparison operators return a size_t which describes the magnitude of the relative affinity and having the comparison operators return a boolean.

PR #80, opened 2018-10-07
Issue #41: Separate execution and memory resources

### LINKED ISSUE #41: Affinity asymmetry between execution and memory resource

AerialMantis:
* Fixes #41 and #42.
* Introduce `memory_resource` to represent the memory component of a system topology.
* Remove `can_place_memory` and `can_place_agents` from the `execution_resource` as these are no longer required.
* Remove `memory_resource` and `allocator` from the `execution_context` as these no longer make sense.
* Update the wording to describe how execution resources and memory resources are structured.
* Refactor `affinity_query` to be between an `execution_resource` and a `memory_resource`.
* Make minor corrections.


affinity/cpp-20/d0796r3.md
```diff
@@ -311,34 +329,44 @@ int main(int argc, char * argv[]) {
...
+auto relativeLatency0 = execution::affinity_query<execution::affinity_operation::read,
+  execution::affinity_metric::latency>(systemResource[0], systemResource.memory_resource());
 
-auto relativeLatency02 = execution::affinity_query<execution::affinity_operation::read,
-  execution::affinity_metric::latency>(systemResource[0], systemResource[2]);
+auto relativeLatency1 = execution::affinity_query<execution::affinity_operation::read,
+  execution::affinity_metric::latency>(systemResource[1], systemResource.memory_resource());
 
-auto relativeLatency = relativeLatency01 > relativeLatency02;
+auto relativeLatency = relativeLatency1 > relativeLatency0;
```
t-tye: What are the requirements for the returned type of the comparison operators? Must they be bool?

affinity/cpp-20/d0796r3.md
```diff
@@ -348,31 +376,32 @@ execution::execution_context execContext(systemResource[0]);
...
+pmr::polymorphic_allocator<int> alloc{socket.memory_resource()};
+
+pmr::vector<int> vec(100, alloc);
+std::generate(par.on(executor), std::begin(vec), std::end(vec), std::rand);
 ```
-*Listing 6: Example of allocating with affinity to an execution resource*
+*Listing 6: Example of executing and allocating with affinity*
 
-The construction of an `execution_context` on a component implies affinity (where possible) to the given resource. This guarantees that all executors created from that `execution_context` can access the resources and the internal data structures requires to guarantee the placement of the processor.
+The construction of an `execution_context` on an `execution_resource` implies affinity (where possible) to the given resource. This guarantees that all executors created from that `execution_context` can access the resources and the internal data structures requires to guarantee the placement of the processor.
```
t-tye: "data structures requires" -> "data structures required"

Not sure what "guarantee the placement of the processor" means. Has "processor" been defined in this context?

affinity/cpp-20/d0796r3.md
```diff
@@ -348,31 +376,32 @@ execution::execution_context execContext(systemResource[0]);
...
 
-Only developers that care about resource placement need to care about obtaining executors and allocations from the correct `execution_context` object. Existing code for vectors and STL (including the Parallel STL interface) remains unaffected.
+Only developers that care about resource placement need to care about obtaining executorsfrom the correct `execution_context` object. Existing code for vectors and STL (including the Parallel STL interface) remains unaffected.
```
t-tye: "executorsfrom" -> "executors from"

affinity/cpp-20/d0796r3.md
```diff
@@ -588,13 +640,66 @@ iterator_traits<>iterator_category
 
 *Returns:* An implementation defined string.
 
-    bool can_place_memory() const noexcept;
+## Class `memory_resource`
+
+The `memory_resource` class provides an abstraction which represents a **memory resource**, that can allocate memory. A `memory_resource` can represent further `memory_resource`s. We say that these `ememory_resource`s are *members of* this `memory_resource`.
```
t-tye: "`ememory_resource`s"->"`memory_resource`s"

affinity/cpp-20/d0796r3.md
mhoemmen: Classes like `less<T>` break if the comparison operators don't return `bool`, because their `operator()` expects to return `bool`.  This issue came up with bitmasks in the SIMD proposal; the author avoided the issue by not overloading comparison operators to return non-`bool`, and instead introducing two-letter comparison operators that return a bitmask for use e.g., in `where` clauses.

mhoemmen: See e.g., *[comparisons.less]*.  I also read *[expr.rel]* as requiring that `operator<` always return `bool`.

affinity/cpp-20/d0796r3.md
```diff
@@ -348,31 +376,32 @@ execution::execution_context execContext(systemResource[0]);
...
+
+/* Create an execution_context and executor associated with the CPU socket */
+execution_context context{socket}
+auto executor = context.executor();
+
+/* Create an allocator from the memory resource associate with the GPU socket */
+pmr::polymorphic_allocator<int> alloc{socket.memory_resource()};
+
+pmr::vector<int> vec(100, alloc);
+std::generate(par.on(executor), std::begin(vec), std::end(vec), std::rand);
```
mhoemmen: `std::rand` is not thread safe, so it's not so good to use for the example here.

affinity/cpp-20/d0796r3.md
mhoemmen: "placement of the data" perhaps?

affinity/cpp-20/d0796r3.md
```diff
@@ -496,7 +550,7 @@ The `execution_resource` which underlies the current thread of execution can be
     /* This thread */
 
     namespace this_thread {
-      std::experimental::execution::execution_resource get_resource() noexcept;
+      const  std::experimental::execution::execution_resource get_resource() noexcept;
```
mhoemmen: What's the point of returning a `const` value here?

affinity/cpp-20/d0796r3.md
```diff
@@ -539,8 +593,6 @@ The `execution_resource` class provides an abstraction over a system's hardware,
 
 *Requires:* `const_iterator` to model `RandomAccessIterator` with the value type `execution_resource::value_type`.
```
mhoemmen: Does "with the value type X" mean "`is_same_v<iterator_traits<const_iterator>::value_type, X>` is `true`"?

affinity/cpp-20/d0796r3.md
```diff
@@ -588,13 +640,66 @@ iterator_traits<>iterator_category
...
 
-    bool can_place_agent() const noexcept;
+> [*Note:* Creating an `memory_resource` may require initializing the  underlying software abstraction when the `memory_resource` is constructed, in order to discover other `memory_resource`s accessible through it. However, an `memory_resource` is nonowning. *--end note*]
 
-*Returns:* If the associated **execution resource* is capable of execute with affinity, 'true'.
+### `memory_resource` member types
+
+    iterator
+
+*Requires:* `iterator` to model `RandomAccessIterator` with the value type `memory_resource::value_type`.
```
mhoemmen: See my comment above.

affinity/cpp-20/d0796r3.md
```diff
@@ -588,13 +640,66 @@ iterator_traits<>iterator_category
...
+
+    const_iterator
+
+*Requires:* `const_iterator` to model `RandomAccessIterator` with the value type `memory_resource::value_type`.
```
mhoemmen: See my comment above.

affinity/cpp-20/d0796r3.md
```diff
@@ -715,7 +798,7 @@ The free function `this_system::discover_topology` is provided for discovering t
 
 The free function `this_thread::get_resource` is provided for retrieving the `execution_resource` underlying the current thread of execution.
 
-    std::experimental::execution::execution_resource get_resource() noexcept;
+    const std::experimental::execution::execution_resource get_resource() noexcept;
```
mhoemmen: Why return a `const` value?

affinity/cpp-20/d0796r3.md
mhoemmen: Also, I think the preferred way to say this is "`const_iterator` satisfies the *Cpp17RandomAccessIterator* requirements."

affinity/cpp-20/d0796r3.md
```diff
@@ -311,34 +329,44 @@ int main(int argc, char * argv[]) {
...
+The `affinity_query` class template provides an abstraction of the relative affinity between an `execution_resource` and a `memory_resource` for a particular `affinity_operation` and `affinity_metric`. The `affinity_query` takes the `affinity_operation` and `affinity_metric` as template parameters, and is constructed from an `execution_resource` and a `memory_resource`.
+
+An `affinity_query` is not generally not meaningful on its own. Instead, users are meant to compare two `affinity_query`s via comparison operators, in order to get a relative magnitude of affinity. If necessary, the underlying value of an `affinity_query` can be queried through `native_affinity`, though the return value of this is implementation defined.
```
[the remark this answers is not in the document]
AerialMantis: Fixed.

affinity/cpp-20/d0796r3.md
AerialMantis: Good catch, I believe this wording is describing the guarantee that executors can make about where execution agents run. I've changed this to "binding of execution agents", does this read better?

affinity/cpp-20/d0796r3.md
AerialMantis: Fixed.

affinity/cpp-20/d0796r3.md
```diff
@@ -278,29 +284,41 @@ Below *(Listing 2)* is an example of executing a parallel task over 8 threads us
...
+![alt text](system-topology.png "Example system topology DAG")
+*Figure 2: Example system topology DAG*
+
 ### Execution resources
 
-An `execution_resource` is a lightweight structure which identifies a particular **execution resource** within a snapshot of the **system topology**. It can be queried for whether the associated **execution resource** can allocate memory via `can_place_memory`, whether the associated **execution resource** can execute work via `can_place_agents`, and for a name via `name`.
+An `execution_resource` is a lightweight structure which identifies a particular **execution resource** within a snapshot of the **system topology**. It can be queried for a name via `name`.
 
-An `execution_resource` object can be queried for a pointer to it's parent `execution_resource` via `member_of`, and can also be iterated over for it's child `execution_resource`s via `begin` and `end`.
+An `execution_resource` object can be queried for a pointer to it's parent `execution_resource` via `member_of`, and can also iterate over it's children `execution_resource`s via `begin` and `end` or access a particular child via `operator[]`.
```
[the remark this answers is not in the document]
AerialMantis: Fixed.

affinity/cpp-20/d0796r3.md
AerialMantis: Fixed.

affinity/cpp-20/d0796r3.md
```diff
@@ -278,29 +284,41 @@ Below *(Listing 2)* is an example of executing a parallel task over 8 threads us
...
 
 An `execution_resource` object can also be queried for the amount concurrency it can provide, the total number of **threads of execution** supported by the associated **execution resource**.
 
-> [*Note:* An **execution resource** is not limited to resources which execute work, but also a general resource where no execution can take place but memory can be allocated, such as off-chip memory. *--end note*]
+An `execution_resource` object can be queried for a pointer to the root `memory_resource` representing the **memory resource** which the associated **execution resource** can access via `memory_resource`.
+
+> [*Note:* There may be a number of **memory resources** which an **execution resource** can access, but the `memory_resource` pointer returned from `memory_resource` should represent the most course grained of these. We may want to expand on this interface in the future. *--end note*]
```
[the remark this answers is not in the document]
AerialMantis: Fixed. I always make that mistake :)

affinity/cpp-20/d0796r3.md
```diff
@@ -311,34 +329,44 @@ int main(int argc, char * argv[]) {
...
 }
 ```
-*Listing 3: Example of querying all the system level execution resources*
+*Listing 3: Example of printing out all execution resources*
+
+### Memory resources
+
+An `memory_resource` is a lightweight structure which inherits from `pmr::memory_resource` and identifies a particular **memory resource** within a snapshot of the **system topology**. It can be queried for a name via `name`.
+
+A `memory_resource` object can be queried for a pointer to it's parent `memory_resource` via `member_of`, and can also iterate over it's children `memory_resource`s via `begin` and `end` or access a particular child via `operator[]`.
```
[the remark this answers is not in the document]
AerialMantis: Fixed.

affinity/cpp-20/d0796r3.md
```diff
@@ -348,31 +376,32 @@ execution::execution_context execContext(systemResource[0]);
...
-auto executor = eC.executor(); // By transitivity, associated with the socket too
-auto socketAllocator = eC.allocator(); // Retrieve an allocator to the closest memory node
-std::vector<int, decltype(socketAllocator)> v1(100, socketAllocator);
-std::generate(par.on(executor), std::begin(v1), std::end(v1), std::rand);
+
+/* find_socket_resource is a user-defined function that finds a resource that is a CPU socket in the given resource list */
+auto socket = find_socket_resource(systemResource);
+
+/* Create an execution_context and executor associated with the CPU socket */
+execution_context context{socket}
```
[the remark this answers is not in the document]
AerialMantis: Fixed.

affinity/cpp-20/d0796r3.md
AerialMantis: The return type of the comparison operators is currently a `std::expected<size_t, error_type>`, the reason for this was so that we could reflect a magnitude rather than simply a boolean as the relative affinity, and also that we want to handle failures to retrieve affinity. However, there have been some concerns about this interface not playing well with others, and this is another good example of that. I have an issue open for looking into alternative solutions to this - https://github.com/codeplaysoftware/standards-proposals/issues/68.

affinity/cpp-20/d0796r3.md
AerialMantis: That's a good point, I've replaced `std::rand` with a generic `genFunc` for now, but we should try to improve the motivational examples in the future.

AerialMantis: The reason this is `pmr::vector` is because we're constructing the vector using a `pmr::polynmorphic_allocator`. It could also be a regular `std::vector` with the `polymorphic_allocator` specified explicitly.

affinity/cpp-20/d0796r3.md
AerialMantis: Yes, that's right, I've made the wording more explicit here, and also took your suggestion for the random access iterator requirements.

affinity/cpp-20/d0796r3.md
AerialMantis: Fixed.

affinity/cpp-20/d0796r3.md
AerialMantis: Fixed.

PR #83, opened 2018-12-03
CP019: Initial draft of on-chip memory proposal

AerialMantis:
SYCL vendor extension proposal to introduce a property to buffers for allocating on-chip memory.


keryell:
That seems like a good idea.

PR #86, opened 2019-01-13
D0796r4 paper split

AerialMantis:
This PR creates two new documents:
* DXXXXr0: Executor properties for affinity-based execution
* DXXXXr0: System topology discovery for heterogeneous & distributed computing

The content from the current D0796r4 has been distributed across the two new documents as follows:
* Everything describing the high-level executor properties and the concept of an execution resource have been moved to DXXXXr0: Executor properties for affinity-based execution.
* Everything describing topology discovery, execution resource lifetimes and the `execution_resource`, `execution_context`, `affinity_query` classes and `this_system::discover_topology` function have been moved to DXXXXr0: System topology discovery for heterogeneous & distributed computing.

There is no new content in this PR except from some altering of the abstract sections.

I still plan to do some further cleanup, including altering the problem space sections to better reflect the new distribution of content.


affinity/cpp-20/dXXX1r0.md
```diff
@@ -0,0 +1,373 @@
...
+```cpp
+// C++ valarray STL containers are initialized automatically.
+// First-touch allocation thus places all of a on the master.
+std::valarray<double> a(N);
+
+// Data placement is wrong, so parallel update is slow.
+std::for_each(std::execution::par, std::begin(a), std::end(a), 
+              [=] (double& a_i) { a_i *= scalar; });
+	      
+// Use future affinity interface to migrate data at next
```
mhoemmen: Per the call today (Mon 14 Jan 2019), we suggested keeping this example (perhaps changing valarray to vector), and showing the interface that would migrate pages of an existing allocation.  We also suggested adding an example that shows coupled control of allocation and execution -- e.g., "scatter."

AerialMantis: Thanks for the suggestion, since this PR is simply re-structuring I will merge this PR and open another with the new examples based on the discussion on today's call.

PR #87, opened 2019-01-16
CP013: Introduce new properties and examples to affinity paper

AerialMantis:
**Changes**

* Replace current motivational example with two new examples: one
showing migration of a pre-allocated container to gain affinity, and one
showing using the bulk_execution_affinity property to bind allocation
and execution locality.
* Introduce the `concurrency_t property`, reflecting that which was
previously available via the `execution_context`.
* Introduce the `execution_locality_intersection_t` property which
returns the maximum concurrency common to two executors.
* Introduce the `memory_locality_intersection_t` property which returns
true if two executors have a common address space.
* Remove sections of background research not relative to high-level
interface.
* Add examples for each of the new properties.
* Update future work section to be more relative to the high-level
interface.
* Make some other minor editorial fixes.

### DIFF
--- a/affinity/cpp-20/d1436r0.md
+++ b/affinity/cpp-20/d1436r0.md
@@ -22,6 +22,7 @@
 * Introduce new executor property `concurrency_t`.
 * Introduce new executor property `execution_locality_intersection_t`.
 * Introduce new executor property `memory_locality_intersection_t`.
+* Update direction for future work.
 
 ### P0796r3 (SAN 2018)
 
@@ -58,11 +59,9 @@
 
 # Abstract
 
-This paper is the result of a request from SG1 at the 2018 San Diego meeting to split P0796: Supporting Heterogeneous & Distributed Computing Through Affinity [[35]][p0796] into two separate papers, one for the high-level interface and one for the low-level interface. This paper focusses on the high-level interface: a series of properties for querying affinity relationships and requesting affinity on work being executed. [[36]][pXXXX] focusses on the low-level interface: a mechanism for discovering the topology and affinity properties of a given system.
+This paper is the result of a request from SG1 at the 2018 San Diego meeting to split P0796: Supporting Heterogeneous & Distributed Computing Through Affinity [[35]][p0796] into two separate papers, one for the high-level interface and one for the low-level interface. This paper focusses on the high-level interface: a series of properties for querying affinity relationships and requesting affinity on work being executed. [[36]][p1437] focusses on the low-level interface: a mechanism for discovering the topology and affinity properties of a given system.
 
-This paper provides an initial meta-framework for the drives toward an execution and memory affinity model for C\+\+.  It accounts for feedback from the Toronto 2017 SG1 meeting on Data Movement in C\+\+ [[1]][p0687r0] that we should define affinity for C\+\+ first, before considering inaccessible memory as a solution to the separate memory problem towards supporting heterogeneous and distributed computing.
-
-This paper proposes a series of executor properties which can be used to apply affinity requirements to bulk execution functions.
+The aim of this paper is to provide a number of executor properties that if supported allow the user of an executor to query and manipulate the binding of *execution agents* and the underlying *execution resources* of the *threads of execution* they are run on.
 
 
 # Motivation
@@ -79,7 +78,13 @@ The affinity problem is especially challenging for applications whose behavior c
 
 Frequently, data are initialized at the beginning of the program by the initial thread and are used by multiple threads. While some OSes automatically migrate threads or data for better affinity, migration may have high overhead. In an optimal case, the OS may automatically detect which thread access which data most frequently, or it may replicate data which are read by multiple threads, or migrate data which are modified and used by threads residing on remote locality groups. However, the OS often does a reasonable job, if the machine is not overloaded, if the application carefully uses first-touch allocation, and if the program does not change its behavior with respect to locality.
 
-Consider a code example *(Listing 1)* that uses the C\+\+17 parallel STL algorithm `for_each` to modify the entries of an `std::vector` `data`.  The example applies a loop body in a lambda to each entry of the `std::vector` `data`, using an execution policy that distributes work in parallel across multiple CPU cores. We might expect this to be fast, but since `std::vector` containers are initialized automatically and automatically allocated on the master thread's memory, we find that it is actually quite slow even when we have more than one thread.
+The affinity interface we propose should help computers achieve a much higher fraction of peak memory bandwidth when using parallel algorithms. In the future, we plan to extend this to heterogeneous and distributed computing. This follows the lead of OpenMP [[2]][design-of-openmp], which has plans to integrate its affinity model with its heterogeneous model [3]. (One of the authors of this document participated in the design of OpenMP's affinity model.)
+
+## Motivational examples
+
+To identify the requirements for supporting affinity we have looked at a number of use cases where affinity between memory locality and execution can provide better performance.
+
+Consider the following code example *(Listing 1)* where the C\+\+17 parallel STL algorithm `for_each` is used to modify the elements of a `std::vector` `data` on an *executor* that will execute on a NUMA system with a number of CPU cores. However the memory is allocated by the `std::vector` default allocator immediately during the construction of `data` on memory local to the calling *thread of execution*. This means the memory allocated for `data` has bad locality to the NUMA regions on the system, therefore as is would incur higher latency when accessed. In this example, this is avoided by migrating `data` to have better affinity with the NUMA regions on the system using an *executor* with the `bulk_execution_affinity.scatter` property applied, before it is accessed by the `for_each`. Note that a mechanism for miration is not yet specified in this paper so this is example currently uses a vendor API; `vendor_api::migrate`, though the intention is that this will be replaced in a future revision.
 
 ```cpp
 // NUMA executor representing N NUMA regions.
@@ -90,13 +95,13 @@ numa_executor exec;
 std::vector<float> data(N * SIZE);
 
 // Require the NUMA executor to bind its migration of memory to the underlying
-// memory resources in a scatter pattern.s
+// memory resources in a scatter pattern.
 auto affinityExec = std::execution::require(exec,
   bulk_execution_affinity.scatter);
 
 // Migrate the memory allocated for the vector across the NUMA regions in a
 // scatter pattern.
-std::execution::migrate(std::begin(data), std::end(data), affinityExec);
+vendor_api::migrate(std::begin(data), std::end(data), affinityExec);
 
 // Placement of data is local to NUMA region 0, so data for execution on other
 // NUMA nodes must is migrated when accessed.
@@ -104,6 +109,8 @@ std::for_each(std::execution::par.on(affinityExec), std::begin(data),
   std::end(data), [=](float &value) { do_something(value): });
 ```
 *Listing 1: Migrating previously allocated memory.*
+
+Now consider a similar code example *(Listing 2)* where again the C\+\+17 parallel STL algorithm `for_each` is used to modify the elements of a `std::vector` `data` on an *executor* that will execute on a NUMA system with a number of CPU cores. However instead of migrating `data` to have affinity with the NUMA regions, `data` is allocated within a bulk execution by an *executor* with the `bulk_execution_affinity.scatter` property applied so that `data` is allocated with affinity. Then when the `for_each` is called with the same executor, `data` maintains its affinity with the NUMA regions.
 
 ```cpp
 // NUMA executor representing N NUMA regions.
@@ -131,11 +138,20 @@ std::for_each(std::execution::par.on(affinityExec), std::begin(data),
 ```
 *Listing 2: Aligning memory and process affinity.*
 
-The affinity interface we propose should help computers achieve a much higher fraction of peak memory bandwidth when using parallel algorithms. In the future, we plan to extend this to heterogeneous and distributed computing. This follows the lead of OpenMP [[2]][design-of-openmp], which has plans to integrate its affinity model with its heterogeneous model [3]. (One of the authors of this document participated in the design of OpenMP's affinity model.)
 
-# Background Research: State of the Art
+# Background Research
 
-The problem of effectively partitioning a system’s topology has existed for some time, and there are a range of third-party libraries and standards which provide APIs to solve the problem. In order to standardize this process for C\+\+, we must carefully look at all of these approaches and identify which we wish to adopt. Below is a list of the libraries and standards from which this proposal will draw:
+In this paper we describe the problem space of affinity for C\+\+, the various challenges which need to be addressed in defining a partitioning and affinity interface for C\+\+, and some suggested solutions. These include:
+
+* How to migrate memory work and memory allocations between execution resources.
+* How to query affinity properties between different *executors*.
+* How to bind execution and allocation particular *executors*.
+
+Wherever possible, we also evaluate how an affinity-based solution could be scaled to support both distributed and heterogeneous systems.
+
+## State of the art
+
+The *affinity problem* existed for some time, and there are a number of third-party libraries and standards which provide APIs to solve the problem. In order to standardize this process for C\+\+, we must carefully look at all of these approaches and identify which we wish ideas are suitable for adoption into C\+\+. Below is a list of the libraries and standards from which this proposal will draw:
 
 * Portable Hardware Locality [[4]][hwloc]
 * SYCL 1.2 [[5]][sycl-1-2-1]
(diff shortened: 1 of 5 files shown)


affinity/cpp-20/dXXX1r0.md
```diff
@@ -54,6 +54,8 @@
 
 # Abstract
 
+This paper is the result of a request from SG1 at the 2018 San Diego meeting to split P0796: Supporting Heterogeneous & Distributed Computing Through Affinity [[35]][p0796] into two separate papers, one for the high-level interface and one for the low-level interface. This paper focusses on the high-level interface; a series of properties for querying affinity relationships and requesting affinity on work being executed. [[36]][pXXXX] focusses on the low-level interface; a mechanism for discovering the topology and affinity properties of a given system.
```
mhoemmen: Use colon instead of semicolon for "focusses on the ... interface: a ...".  (In general, use a colon when the right-hand side of the colon defines the immediate left-hand side.)

affinity/cpp-20/dXXX1r0.md
```diff
@@ -73,25 +75,57 @@ The affinity problem is especially challenging for applications whose behavior c
...
 
-Consider a code example *(Listing 1)* that uses the C\+\+17 parallel STL algorithm `for_each` to modify the entries of a `valarray` `a`.  The example applies a loop body in a lambda to each entry of the `valarray` `a`, using an execution policy that distributes work in parallel across multiple CPU cores. We might expect this to be fast, but since `valarray` containers are initialized automatically and automatically allocated on the master thread's memory, we find that it is actually quite slow even when we have more than one thread.
+Consider a code example *(Listing 1)* that uses the C\+\+17 parallel STL algorithm `for_each` to modify the entries of a `std::vector` `data`.  The example applies a loop body in a lambda to each entry of the `std::vector` `data`, using an execution policy that distributes work in parallel across multiple CPU cores. We might expect this to be fast, but since `std::vector` containers are initialized automatically and automatically allocated on the master thread's memory, we find that it is actually quite slow even when we have more than one thread.
```
mhoemmen: "an `std::vector`"

affinity/cpp-20/dXXX1r0.md
```diff
@@ -73,25 +75,57 @@ The affinity problem is especially challenging for applications whose behavior c
...
+
+```cpp
+// NUMA executor representing N NUMA regions.
+numa_executor exec;
+
+// Storage required for vector allocated on construction local to current thread
+// of execution, (N == 0).
+std::vector<float> data(N * SIZE);
+
+// Require the NUMA executor to bind it's migration of memory to the underlying
```
mhoemmen: "its" (no apostrophe)

affinity/cpp-20/dXXX1r0.md
```diff
@@ -73,25 +75,57 @@ The affinity problem is especially challenging for applications whose behavior c
...
+// memory resources in a scatter patter.
```
mhoemmen: "pattern"

affinity/cpp-20/dXXX1r0.md
```diff
@@ -73,25 +75,57 @@ The affinity problem is especially challenging for applications whose behavior c
...
-// Faster, because data are local now.
-std::for_each(std::execution::par, std::begin(a), std::end(a), 
-              [=] (double& a_i) { a_i *= scalar; });
+// NUMA executor representing N NUMA regions.
+numa_executor exec;
+
+// Reserve space in a vector for a unique_ptr for each index in the bulk
+// execution.
+std::vector<std::unique_ptr<float>> data{};
+data.reserve(N * SIZE);
```
mhoemmen: I'm a bit confused about this example.  I imagine that this data structure would be a vector of size N, with entries of type `unique_ptr<float[]>`, each of which has size `SIZE`.  The comment "[l]aunch a bulk execution that will allocate each `unique_ptr` in the vector with locality to the nearest NUMA region" seems to suggest this.  Also, the array of arrays data structure would match a PGAS-style explicit affinity programming model.  Is that actually what you meant to do here?

affinity/cpp-20/dXXX1r0.md
```diff
@@ -195,58 +266,147 @@ Below *(Listing 2)* is an example of executing a parallel task over 8 threads us
 ```
 *Listing 2: Example of using the bulk_execution_affinity property*
```
mhoemmen: Your changes added a Listing 2 already; should we renumber or remove this one?

affinity/cpp-20/dXXX1r0.md
```diff
@@ -195,58 +266,147 @@ Below *(Listing 2)* is an example of executing a parallel task over 8 threads us
...
 
+### Proposed Wording
 
-## Header `<execution>` synopsis
+The `bulk_execution_affinity_t` property is a behavioral property as defined in P0443 [[22]][p0443] which describes the guarantees an executor provides to the binding of *execution agents* created by a call to `bulk_execute` to the underlying *threads of execution* and to the locality of those *threads of execution*.
+
+The `bulk_execution_affinity_t` property provides nested property types and objects as described below, where:
+* `e` denotes an executor object of type `E`,
+* `f` denotes a function object of type `F&&`,
+* `s` denotes a shape object of type `execution::executor_shape<E>`,
```
mhoemmen: need "and" after the comma, for the third point only

affinity/cpp-20/dXXX1r0.md
```diff
@@ -195,58 +266,147 @@ Below *(Listing 2)* is an example of executing a parallel task over 8 threads us
...
+* `sf` denotes a function object of type `SF`.
+
+| Nested Property Type | Nested Property Name | Requirements |
+|----------------------|----------------------|--------------|
+| bulk_execution_affinity_t::none_t | bulk_execution_affinity_t::none | A call to `e.bulk_execute(f, s, sf)` may or may not bind the created *execution agents* to the underlying *execution resources*. The affinity binding pattern may or may not be consistent across invocations of the executor's bulk execution function.  |
```
mhoemmen: first column's entries need to be in code font

mhoemmen: Requirements for `none` should just be "none."  I get what you're trying to say, but it's a lot of words to say "no requirements."  Also, does "none" mean "I am not specifying the requirements," or does it mean "I don't want affinity at all"?  Those are two different things.

affinity/cpp-20/dXXX1r0.md
```diff
@@ -195,58 +266,147 @@ Below *(Listing 2)* is an example of executing a parallel task over 8 threads us
...
+| bulk_execution_affinity_t::scatter_t | bulk_execution_scatter_t::scatter | A call to `e.bulk_execute(f, s, sf)` must bind the created *execution agents* to the underlying *execution resources* such that they are distributed across the *execution resources* where each *execution agent* far from it's preceding and following *execution agents*. The affinity binding pattern must be consistent across invocations of the executor's bulk execution function. |
```
mhoemmen: "is far from its" -- also, we could just borrow OpenMP language here

affinity/cpp-20/dXXX1r0.md
```diff
@@ -195,58 +266,147 @@ Below *(Listing 2)* is an example of executing a parallel task over 8 threads us
...
+| bulk_execution_affinity_t::compact_t | bulk_execution_compact_t::compact | A call to `e.bulk_execute(f, s, sf)` must bind the created *execution agents* to the underlying *execution resources* such that they are in sequence across the *execution resources* where each *execution agent* close to it's preceding and following *execution agents*. The affinity binding pattern must be consistent across invocations of the executor's bulk execution function. |
```
mhoemmen: "is close to its" -- also, see above

affinity/cpp-20/dXXX1r0.md
```diff
@@ -195,58 +266,147 @@ Below *(Listing 2)* is an example of executing a parallel task over 8 threads us
...
+| bulk_execution_affinity_t::balanced_t | bulk_execution_balanced_t::balanced | A call to `e.bulk_execute(f, s, sf)` must bind the created *execution agents* to the underlying *execution resources* such that they are in sequence and evenly spread across the *execution resources* where each *execution agent* is close to it's preceding and following *execution agents* and all *execution resources* are utilized. The affinity binding pattern must be consistent across invocations of the executor's bulk execution function. |
```
mhoemmen: "is close to its" -- also, see above.  Also, must all execution resources be used?  What if there's not enough bulk work to use all of them?

affinity/cpp-20/dXXX1r0.md
```diff
@@ -195,58 +266,147 @@ Below *(Listing 2)* is an example of executing a parallel task over 8 threads us
...
+
+> [*Note:* The requirements of the `bulk_execution_affinity_t` nested properties do not enforce a specific binding, simply that the binding follows the requirements set out above and that the pattern is consistent across invocations of the bulk execution functions. *--end note*]
+
+> [*Note:* If two *executors* `e1` and `e2` invoke a bulk execution function in order, where `execution::query(e1, execution::context) == query(e2, execution::context)` is `true` and `execution::query(e1, execution::bulk_execution_affinity) == query(e2, execution::bulk_execution_affinity)` is `false`, this will likely result in `e1` binding *execution agents* if necessary to achieve the requested affinity pattern and then `e2` rebinding to achieve the new affinity pattern. *--end note*]
```
mhoemmen: Users not familiar with affinity won't necessarily understand that rebinding costs something.  Adding a few words to the Note could help them.  For example: "Rebinding *execution agents* to *execution resources* may take substantial time and may affect performance of subsequent code."

affinity/cpp-20/dXXX1r0.md
```diff
@@ -195,58 +266,147 @@ Below *(Listing 2)* is an example of executing a parallel task over 8 threads us
...
+This paper currently proposes the `bulk_execution_affinity_t` properties and it's nested properties for allowing an *executor* to make guarantees as to how *execution agents* are bound to the underlying *execution resources*. However providing control at this level may lead to *execution agents* being bound to *execution resources* within a critical path. A possible solution to this is to allow the *execution context* to be configured with `bulk_execution_affinity_t` nested properties, either instead of the *executor* property or in addition. This would allow the binding of *threads of execution* to be performed at the time of the *execution context* creation.
```
mhoemmen: "its" (no apostrophe)

affinity/cpp-20/dXXX1r0.md
```diff
@@ -195,58 +266,147 @@ Below *(Listing 2)* is an example of executing a parallel task over 8 threads us
...
+
+> [*Note:* The requirements of the `bulk_execution_affinity_t` nested properties do not enforce a specific binding, simply that the binding follows the requirements set out above and that the pattern is consistent across invocations of the bulk execution functions. *--end note*]
+
+> [*Note:* If two *executors* `e1` and `e2` invoke a bulk execution function in order, where `execution::query(e1, execution::context) == query(e2, execution::context)` is `true` and `execution::query(e1, execution::bulk_execution_affinity) == query(e2, execution::bulk_execution_affinity)` is `false`, this will likely result in `e1` binding *execution agents* if necessary to achieve the requested affinity pattern and then `e2` rebinding to achieve the new affinity pattern. *--end note*]
+
+> [*Note:* The terms used for the `bulk_execution_affinity_t` nested properties are derived from the OpenMP properties [[33]][openmp-affinity] including the Intel specific balanced affinity binding [[[34]][intel-balanced-affinity] *--end note*]
+
+## Concurrency property
```
mhoemmen: Readers may wonder how this is relevant to the topic of affinity.  For example, the "explicitly manage memory by affinity region" example I proposed above would call for a different number, namely the number of affinity regions N, not the total available concurrency (>> N).

affinity/cpp-20/dXXX1r0.md
AerialMantis: Thanks, fixed.

affinity/cpp-20/dXXX1r0.md
AerialMantis: Fixed.

affinity/cpp-20/dXXX1r0.md
AerialMantis: Fixed.

affinity/cpp-20/dXXX1r0.md
AerialMantis: Fixed.

affinity/cpp-20/dXXX1r0.md
AerialMantis: I forgot to update the listing number, fixed.

affinity/cpp-20/dXXX1r0.md
AerialMantis: Fixed.

affinity/cpp-20/dXXX1r0.md
AerialMantis: Fixed.

affinity/cpp-20/dXXX1r0.md
AerialMantis: Fixed, I also updated the wording for `scatter` to better align with the wording of OpenMP.

affinity/cpp-20/dXXX1r0.md
AerialMantis: Fixed, I also updated the wording for `balanced` to better align with the wording of OpenMP. The way this is now defined the groups of execution resources are sparsely distributed, so this would mean that if there are not enough execution agents in a bulk execution to bind to the available execution resources then the groups are distributed across all of the execution resources, and the execution resources farthest from the start of each group are the ones to not be used.

So for example, 12 execution agents applied to 16 thread execution resources would look something like:
[0][1][3][ ] | [4][5][6][ ] | [7][8][9][ ] | [10][11][12][ ]

I hope this new wording is clearer.

affinity/cpp-20/dXXX1r0.md
AerialMantis: That's a good point, I've included this in the note.

affinity/cpp-20/dXXX1r0.md
AerialMantis: Fixed.

affinity/cpp-20/dXXX1r0.md
AerialMantis: When writting the wording for these properties I found that it's hard to descrbe them (particularly balanced) without refering to the affinity domain (i.e. NUMA region, Core, etc). I'm thinking now this may have to be another property which could potentially be enumerated based on what is available to that particular executor type. What are your thoughts on this?

affinity/cpp-20/dXXX1r0.md
AerialMantis: Yeah, this property isn't specific to affinity, I just wanted to include it because it was previously available to the execution_context in the previous revision.

This is a good point. If we were to introduce an enumeration over different affinity domains, perhaps we could also have a property for querying the breadth of a particular affinity domain?

affinity/cpp-20/dXXX1r0.md
AerialMantis: I've reduced the wording to simply say that there are no requirements on binding.

The intention was that `none` means "I am not specifying any requirements", but you make a good point that this is not the same as no affinity at all. Do you think we need to introduce an additional option? If so perhaps what is currently `none` should be something like `automatic`.

affinity/cpp-20/dXXX1r0.md
mhoemmen: @AerialMantis wrote:
> ... perhaps we could also have a property for querying the breadth of a particular affinity domain?

You would need something like that if you wanted to program explicitly to NUMA domains, as in the "allocate N chunks of memory with explicit affinity" example.  I suspect people will debate about whether that's a "high-level" interface.  It's a bit like calling `omp_get_num_threads` in an OpenMP parallel region instead of just writing a parallel loop in `#pragma omp parallel for` fashion.

affinity/cpp-20/dXXX1r0.md
```diff
@@ -73,25 +75,57 @@ The affinity problem is especially challenging for applications whose behavior c
...
+std::vector<float> data(N * SIZE);
+
+// Require the NUMA executor to bind its migration of memory to the underlying
+// memory resources in a scatter pattern.s
+auto affinityExec = std::execution::require(exec,
+  bulk_execution_affinity.scatter);
+
+// Migrate the memory allocated for the vector across the NUMA regions in a
+// scatter pattern.
+std::execution::migrate(std::begin(data), std::end(data), affinityExec);
```
mhoemmen: I looked at this interface again and wondered: Would it ever make sense to call this for anything other than a contiguous chunk of memory?  For example, could `data` be an `std::map` here?  If not, then maybe this should just take a `span`.

affinity/cpp-20/dXXX1r0.md
```diff
@@ -73,25 +75,57 @@ The affinity problem is especially challenging for applications whose behavior c
...
-              [=] (double& a_i) { a_i *= scalar; });
+// NUMA executor representing N NUMA regions.
+numa_executor exec;
+
+// Reserve space in a vector for a unique_ptr for each index in the bulk
+// execution.
+std::vector<std::unique_ptr<float[SIZE]>> data{};
+data.reserve(N);
+
+// Require the NUMA executor to bind it's allocation of memory to the underlying
```
mhoemmen: "its"

affinity/cpp-20/dXXX1r0.md
```diff
@@ -73,25 +75,57 @@ The affinity problem is especially challenging for applications whose behavior c
...
+auto affinityExec = std::execution::require(exec,
+  bulk_execution_affinity.scatter);
+
+// Launch a bulk execution that will allocate each unique_ptr in the vector with
+// locality to the nearest NUMA region.
+affinityExec.bulk_execute([&](size_t id) {
+  data[id] = std::make_unique<float>(); }, N, sharedFactory);
+
+// Execute a for_each using the same executor so that each unique_ptr in the
+// vector maintains it's locality.
```
mhoemmen: "its"

affinity/cpp-20/dXXX1r0.md
mhoemmen: @AerialMantis `mdspan` is a bit tricky.  Even if you restrict it to 1-D, `mdspan`'s Accessor policy does not require that any raw pointer range actually exist.  Furthermore, migration happens by pages, which could be tricky for a strided `mdspan`.

mhoemmen: @AerialMantis Correction: `mdspan`'s Accessor policy does require that a raw pointer exist somewhere (`a.decay(p)` must return `A::element_type*`), but actually getting a range out isn't necessarily required (hence `A::offset_policy`).

affinity/cpp-20/dXXX1r0.md
AerialMantis: Yeah, I think you're right, it's more of a low-level concept. Though the more I think about it the more I think we might struggle to define the behaviour of these properties without tieing the wording to some flexible concept of an affinity domain. For example, the current properties assume a NUMA like architecture, so may not apply as well to a more heterogeneous architecture.

This is a bit orthogonal to this PR, but I've been thinking an interesting way to represent this could be via a combination of two properties. One for requiring an executor to be structured around a particular affinity domain, i.e. NUMA regions, CPUs, GPUs, cores, threads, etc, that would fail if that domain wasn't valid. And then another for querying or requiring a mask over the resources in the current affinity domain.

For example, you could require an executor to use a NUMA affinity domain, which could have a breadth of say 8 resources (which would be the NUMA regions), and then set the mask to only represent the first half. Or you could have an executor that has both a CPU and a GPU, and you could require it to use a GPU affinity domain, which would enforce that it executes using the GPU.

AerialMantis:
I have finished making changes, I will leave the PR open for a bit longer in case there's any further feedback, then I'll merge it and submit this for the Kona mailing tomorrow.

affinity/cpp-20/d1436r0.md
```diff
@@ -79,7 +78,13 @@ The affinity problem is especially challenging for applications whose behavior c
...
 
-Consider a code example *(Listing 1)* that uses the C\+\+17 parallel STL algorithm `for_each` to modify the entries of an `std::vector` `data`.  The example applies a loop body in a lambda to each entry of the `std::vector` `data`, using an execution policy that distributes work in parallel across multiple CPU cores. We might expect this to be fast, but since `std::vector` containers are initialized automatically and automatically allocated on the master thread's memory, we find that it is actually quite slow even when we have more than one thread.
+The affinity interface we propose should help computers achieve a much higher fraction of peak memory bandwidth when using parallel algorithms. In the future, we plan to extend this to heterogeneous and distributed computing. This follows the lead of OpenMP [[2]][design-of-openmp], which has plans to integrate its affinity model with its heterogeneous model [3]. (One of the authors of this document participated in the design of OpenMP's affinity model.)
```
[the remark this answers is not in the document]
AerialMantis: I've updated this to "integrate its affinity model with its heterogeneity model", does this read better?

affinity/cpp-20/d1436r0.md
```diff
@@ -79,7 +78,13 @@ The affinity problem is especially challenging for applications whose behavior c
...
+
+To identify the requirements for supporting affinity we have looked at a number of use cases where affinity between memory locality and execution can provide better performance.
+
+Consider the following code example *(Listing 1)* where the C\+\+17 parallel STL algorithm `for_each` is used to modify the elements of a `std::vector` `data` on an *executor* that will execute on a NUMA system with a number of CPU cores. However the memory is allocated by the `std::vector` default allocator immediately during the construction of `data` on memory local to the calling *thread of execution*. This means the memory allocated for `data` has bad locality to the NUMA regions on the system, therefore as is would incur higher latency when accessed. In this example, this is avoided by migrating `data` to have better affinity with the NUMA regions on the system using an *executor* with the `bulk_execution_affinity.scatter` property applied, before it is accessed by the `for_each`. Note that a mechanism for miration is not yet specified in this paper so this is example currently uses a vendor API; `vendor_api::migrate`, though the intention is that this will be replaced in a future revision.
```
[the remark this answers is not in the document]
AerialMantis: Good suggestions, I've applied these changes.

affinity/cpp-20/d1436r0.md
```diff
@@ -105,6 +110,8 @@ std::for_each(std::execution::par.on(affinityExec), std::begin(data),
 ```
 *Listing 1: Migrating previously allocated memory.*
 
+Now consider a similar code example *(Listing 2)* where again the C\+\+17 parallel STL algorithm `for_each` is used to modify the elements of a `std::vector` `data` on an *executor* that will execute on a NUMA system with a number of CPU cores. However instead of migrating `data` to have affinity with the NUMA regions, `data` is allocated within a bulk execution by an *executor* with the `bulk_execution_affinity.scatter` property applied so that `data` is allocated with affinity. Then when the `for_each` is called with the same executor, `data` maintains its affinity with the NUMA regions.
```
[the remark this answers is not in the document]
AerialMantis: Fixed.

affinity/cpp-20/d1436r0.md
```diff
@@ -155,64 +171,44 @@ The problem of effectively partitioning a system’s topology has existed for so
...
-* How to query and measure the relative affinity between different execution resources within a system.
-* How to bind execution and allocation particular execution resource(s).
-* What kind of and level of interface(s) should be provided by C\+\+ for affinity.
+This can be scaled to heterogeneous and distributed systems, as the relative affinity between components can apply to discrete heterogeneous and distributed systems as well.
 
-Wherever possible, we also evaluate how an affinity-based solution could be scaled to support both distributed and heterogeneous systems. We also have addressed some aspects of dynamic topology discovery.
+## Inaccessible memory
 
-There are also some additional challenges which we have been investigating but are not yet ready to be included in this paper, and which will be presented in a future paper:
+The initial solution proposed by this paper may only target systems with a single addressable memory region. It may thus exclude devices like discrete GPUs. However, in order to maintain a unified interface going forward, the initial solution should consider these devices and be able to scale to support them in the future.
```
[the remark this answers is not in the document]
AerialMantis: That's a good point, I've reworded this to state that only some discrete GPUs are affected by this.

affinity/cpp-20/d1436r0.md
```diff
@@ -402,23 +428,23 @@ The value returned from `execution::query(e1, memory_locality_intersection_t(e2)
...
 # Future Work
 
-## Who should have control over bulk execution affinity?
+There are a number of additional features which we are considering for inclusion in this paper but are not ready yet.
+
+## Migrating data
+
+This paper currently provides a mechanism for detecting whether two *executors* share a common memory locality. However it does not provide a way to invoke migration of data allocated local to one *executor* into the locality of another *executor*.
+
+We envision that this mechanic could be facilitated by a customization point on two *executors* and perhaps a `spann` or `mdspan` accessor.
```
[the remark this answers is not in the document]
AerialMantis: Good catch, this is what happens when I write papers when I'm tired :)

affinity/cpp-20/d1436r0.md
```diff
@@ -402,23 +428,23 @@ The value returned from `execution::query(e1, memory_locality_intersection_t(e2)
...
 
-| Straw Poll |
-|------------|
-| Should the *execution context* be able to manage the binding of all *threads of execution* which it manages using the `bulk_execution_affinity_t` nested properties? |
-| Should the *executor* be able to manage the binding of all *execution agents* which it manages using the `bulk_execution_affinity_t` nested properties? |
-| Should both the *execution context* and the *executor* be able to manage the binding of *threads of execution* and subsequently *execution agents* using the `bulk_execution_affinity_t` nested properties? |
+This paper currently assumes a NUMA-like system, however there are many other kinds of systems with many different architectures with different kinds of processors, memory and connections between them.
 
-## Migrating data from memory allocated in one partition to another
+In order to accurately take advantage of the range of systems available now and in the future we will need some way to parameterized or enumerate the different affinity domains which an executor can structure around.
```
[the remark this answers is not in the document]
AerialMantis: Fixed.

affinity/cpp-20/d1436r0.md
```diff
@@ -402,23 +428,23 @@ The value returned from `execution::query(e1, memory_locality_intersection_t(e2)
...
 
-With the ability to place memory with affinity comes the ability to define algorithms or memory policies which describe at a higher level how memory is distributed across large systems. Some examples of these are pinned, first touch, and scatter. This is outside the scope of this paper, though we would like to investigate this in a future paper.
+Furthermore in order to have control over those affinity domains we need a way in which to mask out the components of that domain that we wish to work with.
```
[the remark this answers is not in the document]
AerialMantis: Fixed.

affinity/cpp-20/d1436r0.md
```diff
@@ -402,23 +428,23 @@ The value returned from `execution::query(e1, memory_locality_intersection_t(e2)
...
 
-| Straw Poll |
-|------------|
-| Should the interface provide a way of migrating data between partitions? |
+However whichever option we opt for, it must be in such a way as to allow further additions as new system architectures become available.
```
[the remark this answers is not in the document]
AerialMantis: Fixed.

affinity/cpp-20/d1436r0.md
```diff
@@ -131,11 +138,20 @@ std::for_each(std::execution::par.on(affinityExec), std::begin(data),
...
+
+* How to migrate memory work and memory allocations between execution resources.
+* How to query affinity properties between different *executors*.
+* How to bind execution and allocation particular *executors*.
+
+Wherever possible, we also evaluate how an affinity-based solution could be scaled to support both distributed and heterogeneous systems.
+
+## State of the art
+
+The *affinity problem* existed for some time, and there are a number of third-party libraries and standards which provide APIs to solve the problem. In order to standardize this process for C\+\+, we must carefully look at all of these approaches and identify which we wish ideas are suitable for adoption into C\+\+. Below is a list of the libraries and standards from which this proposal will draw:
```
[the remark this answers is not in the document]
AerialMantis: Fixed.

affinity/cpp-20/dXXX1r0.md
AerialMantis: Thanks for looking into this. I think 1-D `mdspan` makes the most sense here, there would have to be a limitation that if the `mdspan` is strided then the stride length must be aligned with the page size (though that's tricky again because AFAIK that's not something available in standard C++). Do you think any further guarantees could be made on an `mdspan` by introducing a specific accessor policy for memory affinity?

Another concern that I had was that for some executors migration can be done by the OS without requiring re-allocation by the user. But in the case of discrete devices such as GPUs that's often not the case. So in this case the migrate operation would have to re-allocate and this means either providing an `mdspan` the ability to arbitrarily reallocate a container or returning a new object. But we can't return a new `mdspan` because as I understand an `mdspan` is simply a view and doesn't own the data.

I think this will require much further discussion, so for now I've made the example simply take the `std::vector` directly to avoid suggesting any particular approach. Does that sound okay?

PR #88, opened 2019-01-21
Add proposal for interop_task

Naghasan:
This proposal add the possibility for a user to retrieve cl_mem object in a friendly way for SYCL data-flow model.

The proposal adds interop_task member function to the codeplay::handler to schedule the execution of a task on the host.
It provides also a way to retrieve cl_mem object associated to an accessor to allow full interoperability with OpenCL.

### DIFF
--- a/interop_task/interop_task.md
+++ b/interop_task/interop_task.md
@@ -126,3 +126,63 @@ class interop_handle {
     };
     qA.submit(cgH);
 ```
+
+This example calls the clFFT library from SYCL using the `interop_task`:
+```cpp
+#include <stdlib.h>
+#include <CL/sycl.hpp>
+
+/* No need to explicitely include the OpenCL headers */
+#include <clFFT.h>
+
+int main( void )
+{
+  size_t N = 16;
+
+  cl::sycl::queue device_queue;
+  cl::sycl::buffer<float> X(range<1>(N * 2));
+
+  /* Setup clFFT. */
+  clfftSetupData fftSetup;
+  err = clfftInitSetupData(&fftSetup);
+  err = clfftSetup(&fftSetup);
+
+  device_queue.submit([=](codeplay::handler& cgh) {
+    auto X_accessor = X.get_access<access::mode::read_write>(cgh);
+    h.interop_task([=](codeplay::interop_handle &handle) {
+      /* FFT library realted declarations */
+      clfftPlanHandle planHandle;
+      size_t clLengths[1] = {N};
+
+      /* Create a default plan for a complex FFT. */
+      err = clfftCreateDefaultPlan(&planHandle, handle.get_context(), CLFFT_1D, clLengths);
+
+      /* Set plan parameters. */
+      err = clfftSetPlanPrecision(planHandle, CLFFT_SINGLE);
+      err = clfftSetLayout(planHandle, CLFFT_COMPLEX_INTERLEAVED, CLFFT_COMPLEX_INTERLEAVED);
+      err = clfftSetResultLocation(planHandle, CLFFT_INPLACE);
+
+      /* Bake the plan. */
+      err = clfftBakePlan(planHandle, 1, &queue, NULL, NULL);
+
+      /* Execute the plan. */
+      cl_command_queue queue = handle.get_queue();
+      cl_mem X_mem = handle.get_buffer(X_accessor);
+      err = clfftEnqueueTransform(planHandle, CLFFT_FORWARD,
+                                  1, &queue, 0, NULL, NULL,
+                                  &X_mem, NULL, NULL);
+
+      /* Wait for calculations to be finished. */
+      err = clFinish(queue);
+
+      /* Release the plan. */
+      err = clfftDestroyPlan( &planHandle );
+    });
+  });
+
+  /* Release clFFT library. */
+  clfftTeardown( );
+
+  return 0;
+}
+```
--- a/interop_task/interop_task.md
+++ b/interop_task/interop_task.md
@@ -12,14 +12,14 @@
 
 ## Motivation
 
-OpenCL applications often build around a set of of fixed function operations which take OpenCL buffers in order to run a hard-coded OpenCL kernels. However, because SYCL does not allow a user to access cl_mem object out of an cl::sycl::accessor, it difficult/impossible to integrate such library into a SYCL application, as the only current way to do this is to create all OpenCL buffers up-front, which is not always possible.
+SYCL does not allow a user to access cl_mem object out of an cl::sycl::accessor, it difficult/impossible to integrate OpenCL library to use them as is inside the data-flow execution model of SYCL, as the only current way to do this is to create all OpenCL buffers up-front, which is not always possible.
 
 This proposal introduces a way for a user to retrieve the OpenCL buffer associate with a SYCL buffer and enqueue a host task that can execute an arbitrary portion of host code within the SYCL runtime, therefore taking advantage of SYCL dependency analysis and scheduling.
 
-## Enqueuing host tasks on SYCL queues
+## Accessing low-level API functionality on SYCL queues
 
 We introduce a new type of handler, the **codeplay::handler**, which includes a new
-**interop\_task** method that executes a task on the host.
+**interop\_task** method that enables submission of low-level API code from the host.
 By submitting this command group to the SYCL device queue, we guarantee it is
 executed in-order w.r.t the other command groups on the same queue.
 Simultaneously, we guarantee that this operation is performed
@@ -28,6 +28,7 @@ thread to continue submitting command groups).
 Other command groups enqueued in the same or different queues
 can be executed following the sequential consistency by guaranteeing the
 satisfaction of the requisites of this command group.
+It is the user's responsability to ensure the lambda submitted via interop_task does not create race conditions with other command groups or with the host.
 
 The possibility of enqueuing host tasks on SYCL queues also enables the
 runtime to perform further optimizations when available.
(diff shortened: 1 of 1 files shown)


interop_task/interop_task.md
```diff
@@ -0,0 +1,123 @@
...
+
+# interop_task: Improving SYCL-OpenCL Interoperability
+
+## Motivation
+
+OpenCL applications often build around a set of of fixed function operations which take OpenCL buffers in order to run a hard-coded OpenCL kernels. However, because SYCL does not allow a user to access cl_mem object out of an cl::sycl::accessor, it difficult/impossible to integrate such library into a SYCL application, as the only current way to do this is to create all OpenCL buffers up-front, which is not always possible.
+
+This proposal introduces a way for a user to retrieve the OpenCL buffer associate with a SYCL buffer and enqueue a host task that can execute an arbitrary portion of host code within the SYCL runtime, therefore taking advantage of SYCL dependency analysis and scheduling.
+
+## Enqueuing host tasks on SYCL queues
```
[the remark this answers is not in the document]
AerialMantis: The terminology here is tricky because it's essentially a hybrid of previous task types. In the context of task scheduling, it is a host task in that it executes a native C++ function within the SYCL scheduler. However, in the context of data-flow the dependencies are on the device side.

Perhaps something like "Enqueuing SYCL interop operations using host tasks"?

interop_task/interop_task.md
```diff
@@ -0,0 +1,123 @@
...
+We introduce a new type of handler, the **codeplay::handler**, which includes a new
+**interop\_task** method that executes a task on the host.
+By submitting this command group to the SYCL device queue, we guarantee it is
+executed in-order w.r.t the other command groups on the same queue.
+Simultaneously, we guarantee that this operation is performed
+asynchronously w.r.t to the user-thread (therefore, enabling the user
+thread to continue submitting command groups).
+Other command groups enqueued in the same or different queues
+can be executed following the sequential consistency by guaranteeing the
+satisfaction of the requisites of this command group.
```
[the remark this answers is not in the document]
AerialMantis: I agree here that it should be the user's responsibility to avoid a race condition with the host or other interop operations, but there's shouldn't be any race conditions between an interop_task and other command groups because the data dependencies have to go through the SYCL runtime.

interop_task/interop_task.md
```diff
@@ -0,0 +1,123 @@
...
+```
+
+### codeplay::handler::interop_task
+
+The `interop_task` allow the user to execute a task of the native host.
+Unlike `single_task`, `parallel_for` and `parallel_for_work_group`, the `interop_task` do not enqueue a kernel on the device but allow the user to execute a custom action when the prerequisites are satisfied on the device associate with the queue.
+The functor passed to the `interop_task` takes as input a const reference to a `cl::sycl::codeplay::interop_handle` which can be used to retrieve underlying OpenCL objects relative to the execution of the task.
+
+It is not allowed to allocate new SYCL object inside an `interop_task`.
+It is the user responsibilities to ensure all asynchronous executions using SYCL provided resources finished before returning from the `interop_task`.
```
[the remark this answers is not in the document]
AerialMantis: I think what this is saying is that the SYCL host API should not be used within an `interop_task`, for example creating a SYCL queue and submitting further command groups.

interop_task/interop_task.md
```diff
@@ -0,0 +1,123 @@
...
+  /* Return the command queue associated with this task */
+  cl_command_queue get_queue() const;
+
+  /*
+    Returns the underlying cl_mem object associated with the accessor
+  */
+  template <typename dataT, int dimensions, access::mode accessmode,
+            access::target accessTarget,
+            access::placeholder isPlaceholder>
+  cl_mem get(const accessor<dataT, dimensions, accessmode, access::target accessTarget, access::placeholder isPlaceholder>&) const;
```
AerialMantis: In SYCL there is a convention to use the name `get` as a way to return the interop object associated with the object calling `get`, so in this case, I think we should use a different name, perhaps `get_buffer`?

interop_task/interop_task.md
```diff
@@ -0,0 +1,123 @@
...
+## Example using regular accessor
+
+```cpp
+    auto cgH = [=] (codeplay::handler& cgh) {
+      auto accA = bufA.get_access<access::mode::read>(cgh);  // Get device accessor to SYCL buffer (cannot be dereference directly in interop_task).
+      auto accB = bufB.get_access<access::mode::read_write>(cgh);
+
+      h.interop_task([=](codeplay::interop_handle &handle) {
+        third_party_api(handle.get_queue(), // Get the OpenCL command queue to use, can be the fallback
+                        handle.get(accA), // Get the OpenCL mem object behind accA
```
AerialMantis: I am not keen on allowing device accessors in an `interop_task` as there's no way to prevent users from accessing the data via `operator[]` which would modify data without the runtime being aware of it, leading to undefined behaviour.

AerialMantis:
Looks good thanks, I think this fills an important gap in the SYCL data-flow model. My only concern here is in allowing device accessors within an `interop_task` as it could make it easier for users to run into undefined behaviour, so I've suggested an improvement to this.

interop_task/interop_task.md
AerialMantis: I think having a new type that is constructed in the command group scope, which performs the parameter marshalling in the interop case would provide the functionality required for interop tasks whilst not allowing users to access the data on the host. Having this be a separate type also makes it easier for compilers or static analysis tools to catch undefined behaviour.

interop_task/interop_task.md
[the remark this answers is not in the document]
Naghasan: changed to `It is the user responsibilities to ensure all operations peroformed inside the interop_task finished before returning from it`

Just to say that the user needs to make sure all operations finished before exiting the lambda.

interop_task/interop_task.md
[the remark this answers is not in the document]
Naghasan: That's a typo inherited from the template file I used, I agree with @Ruyk proposal. We need to emphasis on accessing low level API I think.

interop_task/interop_task.md
Naghasan: I agree with you, @Ruyk pointed out that may force users to modify existing code. Going with the accessor route means less code modification is required for users (they can reuse their data types).

That's a trade-off to find, leaving as it is for now, be revisited later.

interop_task/interop_task.md
```diff
@@ -0,0 +1,124 @@
...
+| Current Status | _Work in progress_ |
+| Reply-to | Victor Lomüller <victor@codeplay.com> |
+| Original author | Victor Lomüller <victor@codeplay.com>, Gordon Brown <gordon@codeplay.com>, Peter Zuzek <peter@codeplay.com> |
+| Contributors | Victor Lomüller <victor@codeplay.com>, Gordon Brown <gordon@codeplay.com>, Peter Zuzek <peter@codeplay.com> |
+
+# interop_task: Improving SYCL-OpenCL Interoperability
+
+## Motivation
+
+SYCL does not allow a user to access cl_mem object out of an cl::sycl::accessor, it difficult/impossible to integrate OpenCL library to use them as is inside the data-flow execution model of SYCL, as the only current way to do this is to create all OpenCL buffers up-front, which is not always possible.
```
Ruyk: it difficult/impossible -> It is difficult to

integrate OpenCL library to use them as is inside -> integrate an OpenCL library seamesly inside the data-flow execution model of SYCL

interop_task/interop_task.md
```diff
@@ -0,0 +1,124 @@
...
+
+This proposal introduces a way for a user to retrieve the OpenCL buffer associate with a SYCL buffer and enqueue a host task that can execute an arbitrary portion of host code within the SYCL runtime, therefore taking advantage of SYCL dependency analysis and scheduling.
```
Ruyk: associate -> associated

interop_task/interop_task.md
```diff
@@ -0,0 +1,124 @@
+| Proposal ID | TBC |
+|-------------|--------|
+| Name |  |
+| Date of Creation | 16 January 2019 |
+| Target | SYCL 1.2.1 extension |
```
Ruyk: *vendor* extension at ths point (hence why it is on the codeplay handler)

interop_task/interop_task.md
```diff
@@ -0,0 +1,124 @@
...
+
+### codeplay::handler::interop_task
+
+The `interop_task` allows users to submit tasks containing C++ statements with low-level API call (e.g. OpenCL Host API entries).
+The command group that encapsulates the task will execute following the usual SYCL dataflow execution rules.
+The functor passed to the `interop_task` takes as input a const reference to a `cl::sycl::codeplay::interop_handle`. The handle can be used to retrieve underlying OpenCL objects relative to the execution of the task.
+
+It is not allowed to allocate new SYCL object inside an `interop_task`.
+It is the user responsibilities to ensure all operations peroformed inside the `interop_task` finished before returning from it.
+
```
Ruyk: Please add:
```
Although the statements inside the lambda submitted o the interop_task are executed on the host, the requirements and actions for the command group are satisied for the device.
This is the opposite of the host_handler vendor extension, where requisites are satisfied for the host since the statements on the lambda submited to the single task are meant to have side effects on the host only.
The interop task lambda can have side effects on the host, but it is the programmer responsability to ensure requirements dont need to be satisfied for the host.
```

interop_task/interop_task.md
Naghasan: I'll replace "OpenCL" by "low-level API functionality" to be consistent with previous changes

PR #89, opened 2019-01-24
Default constructed accessors

ProGTX:
This relates to https://github.com/codeplaysoftware/standards-proposals/pull/18.

It introduces a default constructed placeholder accessor, similar to that patch,
called a null accessor here.
Instead of proposing a `require(buf, acc)`,
the buffer can be "bound" to an accessor by creating a new placeholder accessor
that is already bound to a buffer.
It can then be used as a normal placeholder accessor.
If it's not bound, it can still be passed into a kernel,
but should not be dereferenced inside the kernel.

A new free function `make_placeholder_accessor` accessor is also proposed.

### DIFF
--- a/placeholder_accessors/index.md
+++ b/placeholder_accessors/index.md
@@ -224,7 +224,7 @@ A null accessor does not need to be registered with the command group,
 i.e. there is no need to call `require()` on it.
 It is allowed to pass a null accessor to a kernel,
 as long as it's not dereferenced inside the kernel,
-otherwise there an out-of-bounds access will occur,
+otherwise an out-of-bounds access will occur,
 likely leading to a segmentation fault or similar.
 If the null accessor has been bound to a buffer,
 it is no longer a null accessor,
@@ -236,7 +236,7 @@ by calling the member function `is_null`.
 
 |Member function      |Description                                              |
 |---------------------|---------------------------------------------------------|
-|bool is_null() const |Returns true if the accessor is associated with a buffer, false otherwise.|
+|bool is_null() const |Returns false if the accessor is associated with a buffer, true otherwise.|
 
 
 [1]: https://github.com/codeplaysoftware/sycl-blas "SYCL-BLAS"
--- a/placeholder_accessors/index.md
+++ b/placeholder_accessors/index.md
@@ -9,7 +9,7 @@
 | Reply-to | Ruyman Reyes <ruyman@codeplay.com> |
 | Original author | Ruyman Reyes <ruyman@codeplay.com> |
 | Requirements | CP001 |
-| Contributors | Gordon Brown <gordon@codeplay.com>, Victor Lomuller <victor@codeplay.com>, Mehdi Goli <mehdi.goli@codeplay.com>, Peter Zuzek <peter@codeplay.com>, Luke Iwanski <luke@codeplay.com> |
+| Contributors | Gordon Brown <gordon@codeplay.com>, Victor Lomuller <victor@codeplay.com>, Mehdi Goli <mehdi.goli@codeplay.com>, Peter Žužek <peter@codeplay.com>, Luke Iwanski <luke@codeplay.com>, Michael Haidl <michael.haidl@uni-muenster.de> |
 
 ## Overview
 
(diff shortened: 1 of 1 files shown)


placeholder_accessors/index.md
```diff
@@ -192,6 +193,51 @@ The handler gains a new method,
 that registers the requirements of the placeholder accessor on the given
 command group.
 
+### Placeholder `accessor` without a buffer
+
+Not providing either the handler or the memory object
+makes the accessor default constructible,
+creating a "null" accessor.
```
Ruyk: a "null" accessor -> a "null" or "unbound" placeholder accessor. This represents just the placeholder (i.e. the space that will be occupied by the data type) that needs to be associated with a buffer at command group submission time.

placeholder_accessors/index.md
```diff
@@ -192,6 +193,51 @@ The handler gains a new method,
...
+accessor<T, dim, mode, target, access::placeholder::true_t>
+make_placeholder_accessor(buffer<T, dim, AllocatorT>& bufferRef);
+```
+
+A null accessor does not need to be registered with the command group,
+i.e. there is no need to call `require()` on it.
+It is allowed to pass a null accessor to a kernel,
+as long as it's not dereferenced inside the kernel,
+otherwise an out-of-bounds access will occur,
+likely leading to a segmentation fault or similar.
```
Ruyk: This is quite ambiguous. Is better simply to say a device-specific asynchronous error

placeholder_accessors/index.md
```diff
@@ -192,6 +193,51 @@ The handler gains a new method,
...
+If the null accessor has been bound to a buffer,
+it is no longer a null accessor,
+which means it has to be registered with the command group
+and can be dereferenced inside the kernel.
```
Ruyk: This is also a bit confusing. A null accessor bound to a buffer, is then bound to that buffer forever? can it re-assigned? can it be assigned back to null? What happens if two command groups submit a different require bounding the same null accessor to two separate buffers?

placeholder_accessors/index.md
```diff
@@ -192,6 +193,51 @@ The handler gains a new method,
...
+
+An accessor can be checked for the existence of an associated buffer
+by calling the member function `is_null`.
+
+|Member function      |Description                                              |
+|---------------------|---------------------------------------------------------|
+|bool is_null() const |Returns false if the accessor is associated with a buffer, true otherwise.|
```
Ruyk: What is the expected output of is_null in the following cases?

Case 1:
```cpp
accessor<T, dim, mode, target, access::placeholer::true_t> acc0;

q.submit([&](handler& h) {
      h.require(acc0, bufA);
});

// What is the output here?
std::cout << acc.is_null() << std::endl;
```

Case 2:
```cpp
accessor<T, dim, mode, target, access::placeholer::true_t> acc0;

std::thread t([](){
  q.submit([&](handler& h) {
      h.require(acc0, bufA);
  });
});
// What is the output here?
std::cout << acc.is_null() << std::endl;
t.join();
```

placeholder_accessors/index.md
ProGTX: The main point of this MR is that the `require` function doesn't do the binding - it doesn't modify the accessor at all. So a null accessor can be "bound" by assigning to it a placeholder that already has an associated buffer.

placeholder_accessors/index.md
ProGTX: This PR doesn't have that overload of the `require` function. The existing function doesn't modify the state of the accessor, and having an overload that would modify it is a dangerous inconsistency. That also means that both your cases would print true.
```cpp
h.require(acc0, bufA);
// is instead
acc0 = accessor<T, dim, mode, target, access::placeholer::true_t>(bufA);
h.require(acc0);
// or
acc0 = make_placeholder_accessor<mode, target>(bufA);
h.require(acc0);
```

ProGTX:
This has been replaced by https://github.com/codeplaysoftware/standards-proposals/pull/124, which defines an accessor to be a `Container`, and then you can have "empty" accessors if `acc.empty() == true`.

keryell:
> This has been replaced by #124, which defines an accessor to be a `Container`, and then you can have "empty" accessors if `acc.empty() == true`.

Nice side effect of this container (point-of-) view...

PR #96, opened 2019-08-12
CP013: Make affinity properties hints

AerialMantis:
**Summary**

* Updates the wording for `bulk_execution_affinity` properties to make them hints rather than guarantees.
* Minor editorial fixes to the wording.

**Todo**

* This pull request removes wording that describes guarantees about consistency between invocations of `bulk_execute`, a further property needs to be introduced to reintroduce this functionality.

### DIFF
--- a/affinity/cpp-23/d1436r2.md
+++ b/affinity/cpp-23/d1436r2.md
@@ -15,6 +15,10 @@
 
 # Changelog
 
+### P1436r2 (BEL 2019)
+
+* Alter the wording on the `bulk_execution_affinity_t` properties so they are now hints that request the executor provide a particular pattern of binding, rather than a guarantee.
+
 ### P1436r1 (COL 2019)
 
 * Introduce wording to clarify when two invocations of bulk_execute
@@ -256,7 +260,7 @@ constexpr memory_locality_intersection_t memory_locality_intersection;
 
 ## Bulk execution affinity properties
 
-We propose an executor property group called `bulk_execution_affinity` which contains the nested properties `none`, `balanced`, `scatter` or `compact`. Each of these properties, if applied to an *executor* enforces a particular guarantee of binding *execution agents* to the *execution resources* associated with the *executor* in a particular pattern.
+We propose an executor property group called `bulk_execution_affinity` which contains the nested properties `none`, `balanced`, `scatter` and `compact`. Each of these properties, if applied to an *executor* provides a hint to the `executor` that requests a particular binding of *execution agents* to the *execution resources* associated with the *executor* in a particular pattern.
 
 ### Example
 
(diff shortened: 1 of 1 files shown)


affinity/cpp-23/p1436r1.md
```diff
@@ -293,24 +297,17 @@ The `bulk_execution_affinity_t` property provides nested property types and obje
...
+| bulk_execution_affinity_t::balanced_t | bulk_execution_balanced_t::balanced | A call to `e.bulk_execute(f, s, sf)` should aim to bind the created *execution agents* to the underlying *execution resources* (ordered by physical closeness) such that they are distributed equally across the *execution resources* in a bin packing fashion. <br><br> If the execution context associated with `e` is not able to bind the *execution agents* to the underlying *execution resources* as requested it should fall back to `bulk_execution_affinity_t::none_t`. |
 
-> [*Note:* The requirements of the `bulk_execution_affinity_t` nested properties do not enforce a specific binding, simply that the binding follows the requirements set out above and that the pattern is consistent across invocations of the bulk execution functions. *--end note*]
+> [*Note:* The requirements of the `bulk_execution_affinity_t` nested properties do not enforce any specific binding. *--end note*]
```
mhoemmen: They do, though -- "If the execution context associated with `e` is not able to bind the *execution agents* to the underlying *execution resources* as requested it should fall back to `bulk_execution_affinity_t::none_t`."  Do you think we could delete this Note?

AerialMantis: You're right, that's a good point. I think what I was trying to express here was that an implementation is still free to choose how it maps individual work items to the underlying execution resources, providing it aims to achieve the requested affinity relationship. Would changing to something of that effect make more sense? Otherwise, I am happy to remove the note if it is not particularly helpful.

mhoemmen: @AerialMantis I like that better: "An implementation is free to choose how it maps individual work items to the underlying execution resources, providing it aims to achieve the requested affinity relationship."

AerialMantis: Great thanks! I'll apply the change.

trws:
I'm sorry that I didn't get to this in a reasonable amount of time to discuss it before the mailing, but one thing that  came out of the Cologne discussion was the recommendation that this be worded in terms of assigning work to execution agents rather than agents to resources.  It cuts the amount of new terminology the paper requires, and given a way to have the executor bind the agents (or at least a sensible executor) takes care of the distribution as well, while still giving some freedom to things that don't work that way. That discussion also recommended a new paper to discuss iteration-binning for the algorithms, as you kinda need both for either to be all that effective.  Clearly not for this mailing, but should I put up some issues for these items?

PR #97, opened 2019-08-27
Add proposal for default-constructible buffers

DuncanMcBain:
Default constructible buffers would be a useful addition for generic
code. This proposal adds a zero-argument constructor and functions
for determining whether or not a buffer is usable.

### DIFF
--- a/default-constructed-buffers/default-constructed-buffers.md
+++ b/default-constructed-buffers/default-constructed-buffers.md
@@ -4,10 +4,10 @@ Proposal ID      | TBC
 ---------------- | ---------------------------
 Name             | Default-constructed Buffers
 Date of Creation | 2019-08-20
-Target           |
-Status           | WIP
+Target           | SYCL 1.2.1
+Status           | _Draft_
 Author           | Duncan McBain [Duncan McBain](mailto:duncan@codeplay.com)
-Contributors     |
+Contributors     | Duncan McBain, [Gordon Brown](mailto:gordon@codeplay.com)
 
 ## Description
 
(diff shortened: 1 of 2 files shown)


default-constructed-buffers/default-constructed-buffers.md
```diff
@@ -0,0 +1,88 @@
...
+};
+} // namespace sycl
+} // namespace cl
+```
+The template arguments should remain the same, so that the argument can be
+rebound to a new `buffer` instance later using the copy constructor.
+
+The `is_valid()` call would allow the programmer to query whether or not
+the buffer can be used, i.e. whether or not it was constructed with a range of
+size zero. The operator would call this same function but allow its use in
```
AerialMantis: `operator`  -> `explicit conversion operator`.

default-constructed-buffers/default-constructed-buffers.md
```diff
@@ -0,0 +1,88 @@
...
+
+  explicit operator bool() const noexcept;
+};
+} // namespace sycl
+} // namespace cl
+```
+The template arguments should remain the same, so that the argument can be
+rebound to a new `buffer` instance later using the copy constructor.
+
+The `is_valid()` call would allow the programmer to query whether or not
```
AerialMantis: We should also describe what should happen if you try to use an invalid buffer, i.e. call `get_access`, this should probably throw an exception.

default-constructed-buffers/default-constructed-buffers.md
```diff
@@ -0,0 +1,88 @@
+# Default-constructed Buffers
+
+Proposal ID      | TBC
```
AerialMantis: Remember to add an ID and update the proposal list in the README.

DuncanMcBain:
@AerialMantis I've added some more words that hopefully clarify what default-constructed buffers should do when various of their functions are called. Thanks for the feedback!

PR #100, opened 2019-09-22
Accessor aliases

ProGTX:
In order to reduce the verbosity of programming with SYCL accessors,
this proposal aims to reduce the number of template parameters
to just 2 from the current 4 (placeholders accessors even have 5).
It achieves this by adding aliases to accessors
based on the access target
and slightly revised rules for access modes.

1. Added aliases to accessors
  * `buffer_accessor`,
  * `constant_buffer_accessor
  * `host_accessor`
2. Deduce access mode based on constness of data type
  * `access::mode::read` for `const dataT`
  * `access::mode::read_write` for `dataT`
3. Allow conversions to new aliases
4. Return aliases from buffer access
  * `get_device_access`
  * `get_device_constant_access`
  * `get_host_access`
5. Allow any accessor to be registered with `handler::require`
6. Overload for `handler::require` that also takes an access mode
  * Hint to the scheduler
7. Type traits to help with deducing access modes
  * `access_mode_from_type`
  * `type_from_access_mode`
8. Default all accessor template parameters
  * Expect the data type
  * Default to an accessor to global buffer with read-write access
  * Assumes any accessor can be a placeholder
9. Discussed some considerations and alternatives
10. Examples of reduced verbosity


ProGTX:
> That looks like a good simplification.
> You could also add some CTAD if compiled with C++17 or higher.

@keryell I've greatly simplified the proposal, and I've added a section on CTAD. I think I would need to add deduction guides to the accessor class, I might still add that. There's also the problem that CTAD doesn't work on alias templates in C++17, so maybe I should add something for C++20 as well?

keryell:
That looks good. Probably you need to coordinate with Intel to unify the simplifications they are pushing for too.

accessor-alias/sycl-2.2/index.md
```diff
@@ -0,0 +1,686 @@
...
+in a pointer-like manner.
+However, SYCL ecosystem projects have shown that
+they can be a bit awkward to handle due to their verbosity.
+Storing accessors in function objects requires specifying all template parameters,
+and accessors of different access modes are treated as different types,
+even though there isn't much difference in accessing data.
+
+### Existing `local_accessor` alias
+
+SYCL 1.2.1 already has an alias for local memory,
```
AerialMantis: It might be worth noting here that if this is an alias then you can't use CTAD to further reduce verbosity, perhaps an alternative could be to create a new type, it's a more breaking change, but maybe worth it?

accessor-alias/sycl-2.2/index.md
```diff
@@ -0,0 +1,686 @@
...
+essentially deprecating the `access::placeholder` template parameter.
+This is covered by a separate proposal:
+[CP024 Default placeholders](https://github.com/codeplaysoftware/standards-proposals/pull/122).
+It also doesn't address image accessors.
+We discuss this and some other issues
+in the [Considerations and alternatives](#considerations-and-alternatives) section.
+
+## Defaulting template parameters
+
+We propose defaulting all accessor template parameters
```
AerialMantis: This could conflict with the changes proposed in https://github.com/KhronosGroup/SYCL-Docs/pull/73 to introduce an optional `property_list` to thee `accessor` constructor. Though perhaps some of the information that we currently pass such as the `access::mode` information that is not needed for the type-system could be passed as a property?

accessor-alias/sycl-2.2/index.md
```diff
@@ -0,0 +1,686 @@
...
+    // Allow all combinations
+    const_int_read_write(accIntConst)
+    int_read(accInt)
+    const_int_read(accInt)
+
+    ...
+});
+```
+
+We also propose the following rules for simplifying the access mode:
```
AerialMantis: It could be worth making the `discard` part of the `access::mode` a separate runtime parameter or property?

accessor-alias/sycl-2.2/index.md
ProGTX: I don't quite see how that conflicts, this proposal is about defaulting template parameters, not constructor arguments. Though it's possible some extra deduction guides might be needed because of those changes.

accessor-alias/sycl-2.2/index.md
ProGTX: There's a section `Class template argument deduction` that discusses this, I've added there that `local_accessor` is also affected.

accessor-alias/sycl-2.2/index.md
ProGTX: I was considering that, but it seems like the existing access modes work reasonably well with this proposal, so a new tag might actually be adding verbosity.

ProGTX:
I've pushed revision 0.3, which focuses a lot more on CTAD and makes `host_accessor` a separate type

accessor-alias/sycl-2.2/index.md
ProGTX: I've added `property::discard` and `property::discard_v`

PR #104, opened 2019-10-15
Update default-ctor proposal

DuncanMcBain:
Feedback has been gathered from internal prototyping and we have used
this to inform the proposal.

### DIFF
--- a/default-constructed-buffers/default-constructed-buffers.md
+++ b/default-constructed-buffers/default-constructed-buffers.md
@@ -60,7 +60,7 @@ explicit conversion operator would call this same function but allow its use
 in `if` statements.
 
 Requesting access from a default-constructed buffer should throw an exception.
-It is not meaningful to use a default-constructed buffer on-device. Since there
+It is not meaningful to use a default-constructed buffer on device. Since there
 is no allocation associated with a default-constructed `buffer`,
 `cl::sycl::buffer::set_final_data` and `cl::sycl::buffer::set_write_back`
 should behave as if the `buffer` had a final pointer of `nullptr` at all times.
--- a/default-constructed-buffers/default-constructed-buffers.md
+++ b/default-constructed-buffers/default-constructed-buffers.md
@@ -7,7 +7,7 @@ Date of Creation | 2019-08-20
 Target           | SYCL 1.2.1
 Status           | _Draft_
 Author           | Duncan McBain [Duncan McBain](mailto:duncan@codeplay.com)
-Contributors     | Duncan McBain, [Gordon Brown](mailto:gordon@codeplay.com)
+Contributors     | Duncan McBain, [Gordon Brown](mailto:gordon@codeplay.com), [Ruyman Reyes](mailto:ruyman@codeplay.com)
 
 ## Description
 
@@ -34,7 +34,8 @@ aid their use in this style of generic interface.
 ## Proposal
 
 The `cl::sycl::buffer` class should be augmented with an additional constructor
-that takes no arguments, which initialises the buffer with a zero-size range.
+that takes no arguments, which default-constructs the buffer in an
+implementation-defined manner.
 ```c++
 namespace cl {
 namespace sycl {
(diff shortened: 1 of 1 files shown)


default-constructed-buffers/default-constructed-buffers.md
```diff
@@ -53,18 +54,21 @@ class buffer {
...
+in `if` statements.
 
 Requesting access from a default-constructed buffer should throw an exception.
-It is not meaningful to use a zero-sized allocation on-device. Since there is
-no allocation associated with the `buffer`, `cl::sycl::buffer::set_final_data`
-and `cl::sycl::buffer::set_write_back` should behave as if the `buffer` had a
-final pointer of `nullptr` at all times. The other functions in the `buffer`
-API should behave as though the buffer were constructed with a `range` of size
-zero and otherwise behave normally.
+It is not meaningful to use a default-constructed buffer on-device. Since there
```
ProGTX: Why a hyphen?
```suggestion
It is not meaningful to use a default-constructed buffer on device. Since there
```

DuncanMcBain: Overeager fingers!

default-constructed-buffers/default-constructed-buffers.md
```diff
@@ -53,18 +54,21 @@ class buffer {
...
+in `if` statements.
 
 Requesting access from a default-constructed buffer should throw an exception.
-It is not meaningful to use a zero-sized allocation on-device. Since there is
-no allocation associated with the `buffer`, `cl::sycl::buffer::set_final_data`
-and `cl::sycl::buffer::set_write_back` should behave as if the `buffer` had a
-final pointer of `nullptr` at all times. The other functions in the `buffer`
-API should behave as though the buffer were constructed with a `range` of size
-zero and otherwise behave normally.
+It is not meaningful to use a default-constructed buffer on device. Since there
```
AerialMantis: I think we should strengthen this further to say something to the effect of it's undefined behaviour to use a default constructed buffer (i.e. one for which `has_storage` returns `false`) to create an accessor, either within a SYCL kernel function or on the host via a host accessor.

default-constructed-buffers/default-constructed-buffers.md
```diff
@@ -53,18 +54,21 @@ class buffer {
...
+is no allocation associated with a default-constructed `buffer`,
+`cl::sycl::buffer::set_final_data` and `cl::sycl::buffer::set_write_back`
+should behave as if the `buffer` had a final pointer of `nullptr` at all times.
+The other functions in the `buffer` API should behave as though the buffer were
+constructed with a `range` of size zero and otherwise behave normally.
+
+Explicitly constructing a buffer with a range of zero is not allowed. This does
+not correspond to a valid device allocation.
```
AerialMantis: We should extend this statement to also mention host allocations, where applicable.

default-constructed-buffers/default-constructed-buffers.md
DuncanMcBain: The sentence prior to the one on line 63 here suggests that requesting access from such a buffer should throw an exception. UB to me means many different things *could* happen, whereas an exception being thrown means that (as a library implementer) I can safely request access from any and all buffers passed to me, knowing that the user will get an exception on the other end (through the async handler, presumably). UB on the other hand means that I need to be careful or I'm leaving the program in an unknown state.

What I'm saying is this feels less helpful to the programmer, when we can instead just prescribe that an exception will communicate what went wrong in a "nice" way.

default-constructed-buffers/default-constructed-buffers.md
DuncanMcBain: I guess here I was contrasting the implementation-defined nature of `malloc(0)` compared to the totally-forbidden error from `clCreateBuffer(context, CL_MEM_READ_WRITE, 0ull, nullptr, &err)`. Would changing it to say "This does not correspond to a valid allocation" be an acceptable fix, since it's just flat-out not allowed?

AerialMantis: Yeah, this sounds good to me 👍

default-constructed-buffers/default-constructed-buffers.md
AerialMantis: I thought the intention of the proposal was that it be undefined, so my comment was more suggesting a strengthening of the wording to say this explicitly. But I would actually prefer that this be required to throw an exception so this is fine with me 👍

PR #112, opened 2020-01-04
Update the wording of the bulk_execution_affinity properties in P1436

AerialMantis:
* Rename `bulk_execution_affinity_t::scatter_t` to `bulk_execution_affinity_t::spread_t`.
* Rename `bulk_execution_affinity_t::compact_t` to `bulk_execution_affinity_t::close_t`.
* Refine the wording of the `bulk_execution_affinity_t` properties to clarify the requirements on binding and chunking, based on feedback from SG1.

### DIFF
--- a/affinity/cpp-23/d1436r3.md
+++ b/affinity/cpp-23/d1436r3.md
@@ -18,7 +18,9 @@ This paper is the result of discussions from man contributors within SG1, SG14 a
 
 ### P1436r3 (PRA 2020)
 
-* Refine the wording of the `bulk_execution_affinity_t` properties based on recommendations from SG1.
+* Rename `bulk_execution_affinity_t::scatter_t` to `bulk_execution_affinity_t::spread_t`.
+* Rename `bulk_execution_affinity_t::compact_t` to `bulk_execution_affinity_t::close_t`.
+* Refine the wording of the `bulk_execution_affinity_t` properties to clarify the requirements on binding and chunking, based on feedback from SG1.
 
 ### P1436r2 (BEL 2019)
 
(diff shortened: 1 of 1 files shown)


affinity/cpp-23/d1436r3.md
```diff
@@ -298,18 +302,20 @@ The `bulk_execution_affinity_t` property provides nested property types and obje
...
+| `bulk_execution_affinity_t::none_t` | `bulk_execution_affinity_t::none` | A call to `execution::bulk_execute(e, f, s)` is not required to bind the created *execution agents* for the work-items of the iteration space specified by `s` to *execution resources*. |
+| `bulk_execution_affinity_t::spread_t` | `bulk_execution_spread_t::spread` | A call to `e.bulk_execute(f, s, sf)` should aim to bind the created *execution agents* for the work-items of the iteration space specified by `s` to *execution resources* such that the average locality distance of adjacent work-items in the same subdivision of the available concurrency is maximized and the average locality distance of adjacent work-items in different subdivisions of the available concurrency is maximized. Only binding subsequent *execution agents* to an *execution resource* if no other *execution resources* would otherwise result in fewer *execution agents* being bound to it. <br><br> If `e` is not able to fulfil this aim the it should fall back to `bulk_execution_affinity_t::none_t`. |
```
AerialMantis: Suggest changing:

"Only binding subsequent *execution agents* to an *execution resource* if no other *execution resources* would otherwise result in fewer *execution agents* being bound to it."

to:

"The binding of all *execution agents* to all *execution resources* must not result in the difference between the number of *execution agents* assigned to any *execution resources* being greater than 1."

affinity/cpp-23/d1436r3.md
```diff
@@ -298,18 +302,20 @@ The `bulk_execution_affinity_t` property provides nested property types and obje
 * `s` denotes a shape object of type `execution::executor_shape<E>`, and
 * `sf` denotes a function object of type `SF`.
 
+Where a call to `execution::bulk_execute(e, f, s)` creates a consecutive sequence of work-items from `0` to `s-1`, mapped to the available concurrency of `e`, that is a number of execution resources, which are subdivided in some implementation-defined way.
```
AerialMantis: Add this to the bullet list.

affinity/cpp-23/d1436r3.md
```diff
@@ -298,18 +302,20 @@ The `bulk_execution_affinity_t` property provides nested property types and obje
...
 
-> [*Note:* An implementation is free to choose how it maps individual work items to the underlying *execution resources*, providing it aims to achieve the requested affinity relationship. *--end note*]
+> [*Note:* Note: The subdivision of the available concurrency is implementation-defined. *--end note*]
 
-> [*Note:* It's expected that the default value of `bulk_execution_affinity_t` for most executors be `bulk_execution_affinity_t::none_t`. *--end note*]
+> [*Note:* Note: If the number of work-items specified by `s` is larger than the available concurrency, the manner in which that iteration space is subdivided into a consecutive sequence of work-items is implementation-defined. *--end note*]
```
AerialMantis: Add to this that we will make a follow-up proposal for how to specify chunking patterns.

affinity/cpp-23/d1436r3.md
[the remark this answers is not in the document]
AerialMantis: This has now been updated to the following wording:

> The binding of all execution agents to all execution resources must not result in the difference between the number of execution agents assigned to any execution resources being greater than 1.

I have also added some preamble to introduce the terms of art "locality distance" and "available concurrency":

> The `bulk_execution_affinity_t` nested properties are defined using the following terms of art:
> * *Available concurrency*; which refers to the number of *execution resources* available to an *executor* which can be bound to *execution agents* concurrently, assuming no contention.
> * *Locality distance*; which refers to an implementation-defined metric for measuring the relative affinity between *execution resources* whereby *execution resources* with a lower *locality distance* are likely to have similar latency in memory access operations,  for a given memory location.

> The `bulk_execution_affinity_t` nested properties also refer to the subdivision of *execution resources*, which is an implementation-defined method of subdividing the *available concurrency*, generally based on groupings of *execution resources* with the lowest *locality distance* to each other.

> [*Note:* An alternative term of art for *locality distance* could be *locality interference*. *--end note*]

AerialMantis: Done.

affinity/cpp-23/d1436r3.md
AerialMantis: Done.

affinity/cpp-23/d1436r3.md
AerialMantis: I have added a paragraph to the future work section that reads:

> It is defined in this proposal for the `bulk_execution_affinity_t` properties that when the size in an invocation of `execution::bulk_execute` is greater than the *available concurrency* then it is implementation how that iteration space is subdivided into a consecutive sequence of work-items. The authors of this proposal intend to propose a follow up property for specifying how an iteration space should be subdivided into chunks in this case.

affinity/cpp-23/d1436r3.md
```diff
@@ -139,9 +143,9 @@ std::vector<std::unique_ptr<float[SIZE]>> data{};
 data.reserve(N);
 
 // Require the NUMA executor to bind its allocation of memory to the underlying
-// memory resources in a scatter patter.
+// memory resources in a spread patter.
```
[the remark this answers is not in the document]
AerialMantis: Fixed.

AerialMantis:
@mhoemmen thanks for the feedback.

I have updated the MR to apply the suggested changes, and also added some additional wording to describe the terms of art used in the definition of the `bulk_execution_affinity_t` properties and a note to state that we intend to propose a follow-up property that specifies how larger iteration spaces are subdivided into chunks.

affinity/cpp-23/d1436r3.md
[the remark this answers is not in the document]
mhoemmen: 1. Instead of "which refers to," you could just say "is" -- it's a definition.
2. I like the "available concurrency" definition. :-)
3. "Locality distance" makes sense but lower latency might not necessarily always be the goal.  For example, some kinds of memory might have higher latency for individual accesses, but higher bandwidth for many accesses with a particular access pattern.  The latter makes it the "preferred memory" for some kinds of processors.  Maybe we could fudge this by saying "performance"?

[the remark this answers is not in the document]
AerialMantis: Thanks for the feedback, I've changed the statements as you suggested and I've updated the definition for "locality distance" to also incorporate bandwidth.

mhoemmen:
@AerialMantis Thanks! :-D  Sorry I've been quiet lately -- I've been working on P1673 pretty hard :-) .

affinity/cpp-23/d1436r3.md
```diff
@@ -454,6 +467,10 @@ The value returned from `execution::query(e1, memory_locality_intersection_t(e2)
 
 There are a number of additional features which we are considering for inclusion in this paper but are not ready yet.
 
+## Iteration space subdivision property
+
+It is defined in this proposal for the `bulk_execution_affinity_t` properties that when the size in an invocation of `execution::bulk_execute` is greater than the *available concurrency* then it is implementation how that iteration space is subdivided into a consecutive sequence of work-items. The authors of this proposal intend to propose a follow up property for specifying how an iteration space should be subdivided into chunks in this case.
```
mhoemmen: @AerialMantis "... then it is implementation defined how that iteration space..."

PR #113, opened 2020-01-10
Update proposed direction of P1795

AerialMantis:
* Update the proposed direction to be more specific regarding the approach.
* Introduce a pseudo list of executor properties that are expected to be useful to applications utilizing topology discovery.

### DIFF
--- a/affinity/cpp-23/d1795r2.md
+++ b/affinity/cpp-23/d1795r2.md
@@ -1,6 +1,6 @@
-# P1795r1: System topology discovery for heterogeneous & distributed computing
+# P1795r2: System topology discovery for heterogeneous & distributed computing
 
-**Date: 2019-11-23**
+**Date: 2020-01-10**
 
 **Audience: SG1, SG14**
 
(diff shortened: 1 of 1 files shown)


affinity/cpp-23/d1795r2.md
```diff
@@ -110,29 +113,59 @@ Another example of this is that while Hwloc is highly used in many domains, it n
 
 # 5. Proposed direction
 
+## Overview
+
 This paper aims to build on the unified executors proposal, detailed in P0443 [[14]][p0443r11], so this proposal and any others that stem from it will target P0443 as a baseline, and aim to integrate with its direction as closely as possible.
 
-Below we outline a proposed direction:
+This paper proposes an interface which provides abstract representation of a system's resources and their connections and various properties, capabilities and limitations. This abstract representation will allow a diverse range of architectures both available now and to come in the future to be represented in C++ but without tying down the C++ abstract machine specific hardware definitions such as CPU, GPU, etc.
```
mhoemmen: @AerialMantis sorry I'm a bit late to this -- a minor suggested rewording:

This paper proposes an interface which provides an abstract representation of a system's resources and their connections, and various properties, capabilities and limitations. This abstract representation will allow a diverse range of architectures both available now and to come in the future to be represented in C++, but without tying the C++ abstract machine to specific hardware definitions like "CPU" or "GPU."

PR #118, opened 2020-01-23
CP025: Property list generalization

steffenlarsen:
Adds a proposal for generalizing the use of property lists for all user-constructible SYCL objects.


general_property_lists/general_property_lists.md
```diff
@@ -0,0 +1,239 @@
...
+```cpp
+namespace cl {
+namespace sycl {
+class sampler {
+public:
+    sampler(coordinate_normalization_mode normalizationMode,
+        addressing_mode addressingMode, filterin_mode filteringMode,
+        const property_list &propList = {});
+
+    sampler(cl_sampler clSampler, const context &syclContext,
```
AerialMantis: We tend not to include overloads for the interop constructors with property_list as it's not always possible to enforce the properties.

steffenlarsen: Good point. I have removed this constructor.

PR #125, opened 2020-04-06
Add P2155r0: Policy property for describing adjacency

AerialMantis:
Introduce a draft of the ISO C++ paper for proposing an execution policy property for describing the implications of adjacency in bulk algorithms. This paper is a continuation of P1436, taking in feedback from a discussion in SG1 at the Prague ISO C++ meeting.

Note, the motivation section is taken straight from P1436r2 so I plan to update that to refine it before submission. I also plan to add a changelog that will link back to P1436.

### DIFF
--- a/affinity/cpp-23/dXXXXr0.md
+++ b/affinity/cpp-23/dXXXXr0.md
@@ -1,6 +1,6 @@
 # DXXXXr0: Policy property for describing adjacency
 
-**Date: 2020-03-28**
+**Date: 2020-04-06**
 
 **Audience: SG1, SG14**
 
--- a/affinity/cpp-23/dXXXXr0.md
+++ b/affinity/cpp-23/dXXXXr0.md
@@ -8,7 +8,7 @@
 Rodgers, Mark Hoemmen, Tom Scogland**
 
 **Emails: gordon@codeplay.com, ruyman@codeplay.com, michael@codeplay.com,
-hedwards@nvidia.com, rodgert@twrodgers.com, mhoemme@sandia.gov,
+hedwards@nvidia.com, rodgert@twrodgers.com, mhoemmen@stellarscience.com,
 tscogland@llnl.gov**
 
 **Reply to: gordon@codeplay.com**
@@ -27,41 +27,46 @@ on the the `adjacency` properties, previously named `bulk_execution_affinity`.
 
 # Motivation
 
-*Affinity* refers to the "closeness" in terms of memory access performance,
-between running code, the hardware execution resource on which the code runs,
-and the data that the code accesses.  A hardware execution resource has "more
-affinity" to a part of memory or to some data, if it has lower latency and/or
-higher bandwidth when accessing that memory / those data.
+The typical computer systems targeted by the C++ abstract machine have a uniform
+address space, meaning a pointer can be dereferenced anywhere in the
+application. However, the cost to access different regions of that address space
+are most often not uniform. The position of a memory allocation relative to the
+thread which accesses, otherwise refereed to as the process or memory affinity
+can have a significant impact on the performance of concurrent applications.
 
-On almost all computer architectures, the cost of accessing different data may
-differ. Most computers have caches that are associated with specific processing
-units. If the operating system moves a thread or process from one processing
-unit to another, the thread or process will no longer have data in its new cache
-that it had in its old cache. This may make the next access to those data
-slower. Many computers also have a Non-Uniform Memory Architecture (NUMA),
-which means that even though all processing units see a single memory in terms
-of programming model, different processing units may still have more affinity to
-some parts of memory than others. NUMA exists because it is difficult to scale
-non-NUMA memory systems to the performance needed by today's highly parallel
-computers and applications.
+Most computer systems have numerous level of cache and some larger systems have
+a Non-Uniform Memory Architecture (NUMA), where they have a single address space
+that is virtually mapped across a network of interconnected nodes. NUMA exists
+specifically because it is difficult to scale non-NUMA memory systems to the
+performance needed by today's highly parallel applications.
 
-One strategy to improve applications' performance, given the importance of
-affinity, is processor and memory *binding*. Keeping a process bound to a
-specific thread and local memory region optimizes cache affinity. It also
-reduces context switching and unnecessary scheduler activity. Since memory
-accesses to remote locations incur higher latency and/or lower bandwidth,
-control of thread placement to enforce affinity within parallel applications is
-crucial to fuel all the cores and to exploit the full performance of the memory
-subsystem on NUMA computers. 
+When a memory allocation is shared among threads, you can encounter false
+sharing depending on which threads are reading and writing that location. When
+this happens then the cost of reading from that location can be much higher than
+would ordinarily be expected. This is further complicated as operating systems
+often switch out hardware threads from one thread or process to another.
+
+As the C++ abstract machine has no concept of caches or the positioning of OS
+threads it is impossible for C++ applications to reason about this without using
+a third party programming model such as OpenMP [[3]][design-of-openmp].
+
+One strategy to improve an applications' performance, given the importance of
+affinity, is thread and memory placement. Keeping a process bound to a specific
+thread and local memory region optimizes cache affinity. It also reduces context
+switching and unnecessary scheduler activity. Since memory accesses to remote
+locations incur higher latency and/or lower bandwidth, control of thread
+placement to enforce affinity within parallel applications is crucial to fuel
+all the cores and to exploit the full performance of the memory subsystem on
+NUMA computers. 
 
 Operating systems (OSes) traditionally take responsibility for assigning threads
 or processes to run on processing units. However, OSes may use high-level
 policies for this assignment that do not necessarily match the optimal usage
 pattern for a given application. Application developers must leverage the
-placement of memory and *placement of threads* for best performance on current
+placement of memory and placement of threads for best performance on current
 and future architectures. For C\+\+ developers to achieve this, native support
-for *placement of threads and memory* is critical for application portability.
-We will refer to this as the *affinity problem*. 
+for placement of threads and memory is critical for application portability.
+We will refer to this as the affinity problem. 
 
 The affinity problem is especially challenging for applications whose behavior
 changes over time or is hard to predict, or when different applications
@@ -89,15 +94,16 @@ the lead of OpenMP [[3]][design-of-openmp].
 # Proposal
 
 This paper proposes a new group of properties (as described in P1393r0
-[[4]][p1393r0] that allow users to specify how a callable in a bulk algorithm
-such as the `bulk_execute` algorithm in [[2]][p0443r12] will perform in light of
-the placement of the work-items it invokes. This information can then be used by
-an an executor implementation which is aware of the property to efficiently map
-the work-items it invokes to the underlying resources which it represents.
+[[4]][p1393r0]) that allow users to specify how a callable in a bulk algorithm
+such as the `bulk_execute` algorithm in P0443r12 [[2]][p0443r12] will perform in
+light of the placement of the work-items it invokes. This information can then
+be used by an executor implementation which is aware of the property to
+efficiently map the work-items it invokes to the underlying resources which it
+represents.
 
 ## Shift to descriptive properties
 
-The previous revision of this proposal; P1436r3 [[1]][p1436r3] took a
+The previous revision of this proposal; P1436r3 [[1]][p1436r3], took a
 prescriptive approach. It suggested the `bulk_execution_affinity`
 properties, which targeted executors, providing a hint to the executor
 implementation that it should aim to map the work-items invoked by a bulk
@@ -112,11 +118,11 @@ work-items invoked by a bulk algorithm, allowing the implementation to choose
 the appropriate action.
 
 This shift moves towards a more algorithmic-based approach to the customization
-of execution and both removes the need for users to understand the internal
+of execution. It both removes the need for users to understand the internal
 workings of various executors and allows more freedom in executor
 implementations to choose how to interpret the hint.
 
-The properties of the previous revision of this paper; `bulk_execution_affinity`
+The properties of the previous revision of this paper, `bulk_execution_affinity`
 , also included the `balanced` property which differentiated between round-robin
 and bin-packed mapping of work-items to the underlying resources. This variant
 is not reflected in this paper, as it requires some further investigation, so it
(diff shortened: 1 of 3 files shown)


affinity/cpp-23/dXXXXr0.md
```diff
@@ -0,0 +1,295 @@
...
+
+**Date: 2020-04-06**
+
+**Audience: SG1, SG14**
+
+**Authors: Gordon Brown, Ruyman Reyes, Michael Wong, H. Carter Edwards, Thomas
+Rodgers, Mark Hoemmen, Tom Scogland**
+
+**Emails: gordon@codeplay.com, ruyman@codeplay.com, michael@codeplay.com,
+hedwards@nvidia.com, rodgert@twrodgers.com, mhoemme@sandia.gov,
```
mhoemmen, line 11: Please change my e-mail address to `mhoemmen@stellarscience.com` -- thanks!

affinity/cpp-23/dXXXXr0.md
```diff
@@ -0,0 +1,295 @@
...
+some parts of memory than others. NUMA exists because it is difficult to scale
+non-NUMA memory systems to the performance needed by today's highly parallel
+computers and applications.
+
+One strategy to improve applications' performance, given the importance of
+affinity, is processor and memory *binding*. Keeping a process bound to a
+specific thread and local memory region optimizes cache affinity. It also
+reduces context switching and unnecessary scheduler activity. Since memory
+accesses to remote locations incur higher latency and/or lower bandwidth,
+control of thread placement to enforce affinity within parallel applications is
```
mhoemmen, line 53: "Thread placement" first shows up italicized in the next paragraph, but we use it here.  Do we want to say that placement is the same as binding?

affinity/cpp-23/dXXXXr0.md
```diff
@@ -0,0 +1,295 @@
...
+
+Below is an example of how the `adjacency` properties can be used with the
+`indexed_for` algorithm to perform first touch initialization. For the purposes
+of demonstration this example uses sender-based algorithms from P1897r2
+[[5]][p1897r2].
+
+```cpp
+// Create a vector of unique_ptr and reserve space for one for each index in the
+// bulk algorithm.
+std::vector<std::unique_ptr<float> data{};
```
mhoemmen, line 135: `std::vector<std::unique_ptr<float>> data{};

affinity/cpp-23/dXXXXr0.md
mhoemmen: What I'm getting at, is that threads might get an initial placement, but may or may not be allowed to migrate around from there.  For instance, it could make sense to assign threads initially to certain cores, but permit them to migrate for better load balancing or to avoid overheating.  I see binding as a special case of placement.

affinity/cpp-23/dXXXXr0.md
```diff
@@ -0,0 +1,295 @@
...
+This paper proposes a new group of properties (as described in P1393r0
+[[4]][p1393r0] that allow users to specify how a callable in a bulk algorithm
+such as the `bulk_execute` algorithm in [[2]][p0443r12] will perform in light of
+the placement of the work-items it invokes. This information can then be used by
+an an executor implementation which is aware of the property to efficiently map
+the work-items it invokes to the underlying resources which it represents.
+
+## Shift to descriptive properties
+
+The previous revision of this proposal; P1436r3 [[1]][p1436r3] took a
```
mhoemmen, line 100: "The previous revision of this proposal, P1463r3 [[1]][p1436r3], took a ...."

affinity/cpp-23/dXXXXr0.md
```diff
@@ -0,0 +1,295 @@
...
+pattern to achieve optimal performance for the callable being invoked.
+
+This proposal switches perspective on this, and instead takes a
+descriptive approach. It suggests the `adjacency` properties, which instead
+target execution policies, providing a hint to the executor implementation which
+expresses the performance implications of the callable on the mapping of the
+work-items invoked by a bulk algorithm, allowing the implementation to choose
+the appropriate action.
+
+This shift moves towards a more algorithmic-based approach to the customization
```
mhoemmen, line 114: Consider: "... to the customization of execution.  It both removes...."

affinity/cpp-23/dXXXXr0.md
```diff
@@ -0,0 +1,295 @@
...
+of execution and both removes the need for users to understand the internal
+workings of various executors and allows more freedom in executor
+implementations to choose how to interpret the hint.
+
+The properties of the previous revision of this paper; `bulk_execution_affinity`
```
mhoemmen, line 119: "... of this paper, `bulk_execution_affinity`, ..."

affinity/cpp-23/dXXXXr0.md
```diff
@@ -0,0 +1,295 @@
...
+of abstraction is implementation-defined. An implementation is permitted to
+migrate any underlying resources providing it guarantees the affinity properties
+remain consistent. This allows freedom for the implementor but also consistency
+for users.
+
+If an *execution resource* is valid, then it must always point to the same
+underlying thing. For example, a *resource* cannot first point to one CPU core,
+and then suddenly point to a different CPU core. An *execution context* can thus
+rely on properties like binding of operating system threads to CPU cores.
+However, the "thing" to which an *execution resource* points may be a dynamic,
```
mhoemmen, line 222: "may be a dynamic, possibly software-managed pool of hardware"

affinity/cpp-23/dXXXXr0.md
```diff
@@ -0,0 +1,295 @@
...
+*Locality distance* specifies an implementation-defined metric for
+measuring the relative affinity between *execution resources* whereby *execution
+resources* with a lower *locality distance* are likely to have similar latency
+in memory access operations, for a given memory location.
+
+> [*Note:* For example, two hardware threads on the same CPU core would have a
+lower *locality distance* whereas two hardware threads on different CPU cores or
+in different NUMA regions would have a higher *locality distance*. *--end note*]
+
+> [*Note:* An alternative term of art for *locality distance* could be *locality
```
mhoemmen, line 259: "Distance" suggests comparability, e.g., in terms of minimum distance in a graph.  I think we want that ultimately, especially for introducing the various OpenMP arrangements that refer to different levels of a cache or node hierarchy, so it's not bad to save "distance" for later.

affinity/cpp-23/dXXXXr0.md
AerialMantis: Done.

affinity/cpp-23/dXXXXr0.md
AerialMantis: Fixed.

affinity/cpp-23/dXXXXr0.md
AerialMantis: Done.

affinity/cpp-23/dXXXXr0.md
AerialMantis: Done.

affinity/cpp-23/dXXXXr0.md
AerialMantis: Done.

affinity/cpp-23/dXXXXr0.md
AerialMantis: Good suggestion, done.

affinity/cpp-23/dXXXXr0.md
AerialMantis: That's a good point, do you think we should opt for "locality interference" as the term here?

affinity/cpp-23/dXXXXr0.md
```diff
@@ -0,0 +1,295 @@
...
+
+An *execution resource* represents an abstraction of a hardware or software
+layer that guarantees a particular set of affinity properties, where the level
+of abstraction is implementation-defined. An implementation is permitted to
+migrate any underlying resources providing it guarantees the affinity properties
+remain consistent. This allows freedom for the implementor but also consistency
+for users.
+
+If an *execution resource* is valid, then it must always point to the same
+underlying thing. For example, a *resource* cannot first point to one CPU core,
```
mhoemmen, line 219: Given the VM example below, perhaps we should just delete the "For example..." sentence.  It might be a bit confusing.  The sentence after suffices, I think.

affinity/cpp-23/dXXXXr0.md
mhoemmen: @AerialMantis I think it would be better to use "locality interference."  "Distance" suggests a metric with associated `operator<`, but I don't think we're intending to expose that here.  Even if we intend to expose affinity distances in the other proposal (where you can walk the hardware graph), affinity distances are a different thing than interference.  The former is a hardware property, while the latter is an algorithm property ("does the algorithm benefit from spatial locality?").

Would we ever intend to define a mapping between affinity distance and its benefits to constructive interference?

affinity/cpp-23/dXXXXr0.md
```diff
@@ -0,0 +1,295 @@
...
+in memory access operations, for a given memory location.
+
+> [*Note:* For example, two hardware threads on the same CPU core would have a
+lower *locality distance* whereas two hardware threads on different CPU cores or
+in different NUMA regions would have a higher *locality distance*. *--end note*]
+
+> [*Note:* An alternative term of art for *locality distance* could be *locality
+interference*. *--end note*]
+
+When a bulk algorithm is invoked with a callable `f`, a size `s` and using a
```
AerialMantis, line 262: A thought came to me when writing this, in previous papers we had wording which suggested implementations should aim to map work-items to executions resources in a deterministic way (given the same size). Should we introduce this for the `adjacency` property as a hint or should we leave this as a quality of implementation detail or a future property?

affinity/cpp-23/dXXXXr0.md
AerialMantis: I agree, I've removed this sentence.

affinity/cpp-23/dXXXXr0.md
mhoemmen: @AerialMantis wrote:
> Should we introduce this for the `adjacency` property as a hint or should we leave this as a quality of implementation detail or a future property?

I'd like to see it, not as a quality of implementation detail, but either as default behavior or as something I can request (either a hint or property).  Even if the default is a deterministic mapping, it would be nice to be able to relax that by request.  Existing systems (e.g., Kokkos and OpenMP) have both static (deterministic) and dynamic (nondeterministic) options for assigning chunks of work to execution resources, and I think users would want us to expose that.

AerialMantis: I agree, I think having a property that allows you to specify a level of determinism in the mapping of work-items in bulk algorithms. I can make determinism the default for now as this is most beneficial for the adjacency property.

affinity/cpp-23/dXXXXr0.md
AerialMantis: I agree that locality distance is more of a property of the relationship between components of the topology, so perhaps makes more sense in the topology discovery interface, while locality interference is a property of the algorithm.

Yeah that's a good point, I think as we come to define a vocabulary for describing a system topology it makes sense to relate these to the algorithm properties, so we should keep this in mind.

affinity/cpp-23/dXXXXr0.md
AerialMantis: That's a good point, I've reworked the motivation slightly in my most recent change, to better reflect the new approach. I've tried to focus on the term "thread placement" as this seems like the more general term, while "binding threads" is, as you said, a special case of this.

affinity/cpp-23/dXXXXr0.md
AerialMantis: I've added this as a non-normative note, since the properties are not prescriptive it's hard to require a particular mapping, how does this sound?

Subsequent invocations of a bulk algorithm with the same size `s` and the same policy `p`, scheduling on the same executor `e` should aim to map work-items to *execution resources* consistently with previous invocations.

AerialMantis:
Thanks for the feedback @mhoemmen!

rodgert:
line 34: "thread which accesses, otherwise refereed to"
Should be 'referred'

Should this sentence read - "The position of a memory allocation relative to the thread which accesses the memory, otherwise referred to as memory affinity, can have a significant impact on the performance of concurrent applications." ?

line 37: a nit, "some larger systems" seems to downplay the ubiquitous nature of NUMA architectures, any vanilla two socket server is going to have NUMA these days.

paragraph starting at line 43: this is really rather worse than "simple" false sharing, once we start to talk about NUMA. Not sure if we need to point out the issues with saturating node to node communication links here as well.

line 54: I assume 'process' here means the OS notion of process. Does this even matter, it's the thread affinities within a process that matter, generally. So "keeping a processes thread(s) bound to specific memory regions optimizes" ...

line 58: Is "fuel all the cores" what the kids are calling it these days? Maybe something like "is crucial to maximally utilizing cores and to exploit the full performance of the memory subsystem on NUMA architectures."

line 89: "should help computers" are we helping the computer, or are we helping the programmer help the computer? Maybe this is a nit, but it seems that since we are talking about a programming facility the thing we are helping is the programmer.

AerialMantis:
Thanks very much for the feedback @rodgert! I've applied your suggestions.

> Should this sentence read - "The position of a memory allocation relative to the thread which accesses the memory, otherwise referred to as memory affinity, can have a significant impact on the performance of concurrent applications." ?

Yes, thanks.

> line 37: a nit, "some larger systems" seems to downplay the ubiquitous nature of NUMA architectures, any vanilla two socket server is going to have NUMA these days.

That's a good point, I've fixed this.

> paragraph starting at line 43: this is really rather worse than "simple" false sharing, once we start to talk about NUMA. Not sure if we need to point out the issues with saturating node to node communication links here as well.

That's a good point, I have included that case as well.

> line 54: I assume 'process' here means the OS notion of process. Does this even matter, it's the thread affinities within a process that matter, generally. So "keeping a processes thread(s) bound to specific memory regions optimizes" ...

Yes, that's more accurate, I'll use that.

> line 58: Is "fuel all the cores" what the kids are calling it these days? Maybe something like "is crucial to maximally utilizing cores and to exploit the full performance of the memory subsystem on NUMA architectures."

I like this wording, I'll use this instead.

> line 89: "should help computers" are we helping the computer, or are we helping the programmer help the computer? Maybe this is a nit, but it seems that since we are talking about a programming facility the thing we are helping is the programmer.

That's a good point, I agree that sounds better.

AerialMantis:
Thank you everyone for the feedback, I have now submitted this for the mailing, so I will merge this.

PR #126, opened 2020-04-17
CP28 Automatic placeholder requirements

ProGTX:
1. Allow skipping `require`
2. Define convenience functions
3. Don't allow this on kernel enqueues

### DIFF
--- a/auto-require/index.md
+++ b/auto-require/index.md
@@ -1,0 +1,111 @@
+# Automatic placeholder requirements
+
+| Proposal ID | CP028  |
+|-------------|--------|
+| Name | Automatic placeholder requirements |
+| Date of Creation | 17 April 2020 |
+| Revision | 0.1 |
+| Latest Update | 28 April 2020 |
+| Target | SYCL Next (after 1.2.1) |
+| Current Status | _Work in Progress_ |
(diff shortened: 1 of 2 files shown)


auto-require/index.md
```diff
@@ -0,0 +1,109 @@
...
+## Limitations
+
+There is a temptation to extend this automatic registration
+to member functions of the `handler` that perform kernel enqueues:
+`single_task`, `parallel_for`, `parallel_for_work_group`.
+However, that might prove a bit more difficult
+because the accessors are captured inside a function object
+and C++ doesn't yet provide proper facilities for inspecting that object -
+static reflection would be needed.
+Therefore it is our concern that that feature may not be implementable
```
Ruyk, line 108:
```suggestion
Therefore it is our concern that feature may not be implementable
```

ProGTX: I believe the original wording is correct, but I agree that it's a bit difficult to parse. I've updated it:
```md
Therefore it is our concern that extending the proposal to kernel enqueues
may not be implementable on the SYCL host device.
```

ProGTX:
> Please add a link to the relevant USM and queue simplifications

Done, plus a few minor fixes

AerialMantis (approved):

PR #131, opened 2020-07-02
CP029 Image accessors with format information

steffenlarsen:
Adds an extension for more supported data types for image accessors and a new `get_access` overload on images that allows for image format information to be known by the accessor at compile-time.

### DIFF
--- a/accessor_image_format/accessor_image_format.md
+++ b/accessor_image_format/accessor_image_format.md
@@ -85,6 +85,9 @@ template<> struct image_format_type<image_format::r32g32b32a32_sfloat> {
 template<> struct image_format_type<image_format::b8g8r8a8_unorm> {
     using type = float4;
 };
+
+template<image_format format>
+using image_format_t = typename image_format_type<format>::type;
 } // namespace sycl
 ```
 
@@ -92,7 +95,7 @@ With these, a new overload of `get_access` is added to `image` defined as
 
 ```c++
 template <image_format format, access::mode accessMode>
-accessor<image_format_type<format>::type, dimensions, accessMode, access::target::image>
+accessor<image_format_t<format>, dimensions, accessMode, access::target::image>
 get_access(handler & commandGroupHandler);
 ```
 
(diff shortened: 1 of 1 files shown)


accessor_image_format/accessor_image_format.md
```diff
@@ -0,0 +1,166 @@
...
+};
+template<> struct image_format_type<image_format::r16b16g16a16_sfloat> {
+    using type = half4;
+};
+template<> struct image_format_type<image_format::r32g32b32a32_sfloat> {
+    using type = float4;
+};
+template<> struct image_format_type<image_format::b8g8r8a8_unorm> {
+    using type = float4;
+};
```
ProGTX, line 89: Could we maybe also define a helper alias:

```suggestion
};

template<image_format format>
using image_format_t = typename image_format_type<format>::type;
```

steffenlarsen: Good call. I have added this and changed the data type of the accessor returned by the new `get_accessor` to use this.

I'm not completely convinced that `image_format_type` is the best name for this struct, so I'm open to alternatives.

AerialMantis: Perhaps `image_access_type`?

accessor_image_format/accessor_image_format.md
```diff
@@ -0,0 +1,169 @@
...
+accessor<image_format_t<format>, dimensions, accessMode, access::target::image>
+get_access(handler & commandGroupHandler);
+```
+
+Calling `read` or `write` on an image accessor retrieved through the new
+`get_access` above causes undefined behaviour unless the `image_channel_type` of
+the underlying image and the `image_format` given to `get_access` is one of the
+combinations shown in the following table:
+
+|  `image_channel_type `  |     `image_format`    |
```
AerialMantis: It would be useful to specify the type of `image_format_type_t` for each value of `image_format` as well to save cross-referencing it.

AerialMantis (approved):
Looks good, just some minor comments.

accessor_image_format/accessor_image_format.md
steffenlarsen: I think that's a good name, though I'm not fond of the `_type` suffix. I've renamed it to `image_access`. What do you think?

accessor_image_format/accessor_image_format.md
steffenlarsen: I don't think this would be the table for it, but I could add a new table with the mappings if that would be useful?

PR #134, opened 2021-06-24
CP030 Item view extension

steffenlarsen:
Adds an extension for a new `item_view` SYCL object. The `item_view` allows for expressing a uniform access pattern to device memory, exposing each item in the underlying data as an additional parameter to the kernel function. This feature extends the functionality of variadic `parallel_for` allowing `item_view` objects to be intermingled with reduction objects.


keryell:
Interesting ideas.

Ruyk (approved):

PR #135, opened 2022-09-13
Codeplay graphs vendor extension

EwanC:
Make public the specification for `SYCL_EXT_CODEPLAY_GRAPHS`, Codeplay's vendor extension introducing an API for user capture and submission of command-group graphs.

This is the initial version and still subject to change.


graphs/index.md
```diff
@@ -0,0 +1,706 @@
...
+additionally included in the extension and both interfaces would be available to
+the user, as with [CUDA Graphs][cuda-graphs]. The design of this explicit API
+will be informed by early implementations of the current proposal. This is an
+area we are looking to collaborate with other vendors on towards a KHR extension
+or SYCL Next.
+
+#### Single Queue Submit
+
+This extension allows a graph to be recorded from multiple different queues,
+but ony submitted to a single queue. The device associated with a queue is not
```
reble:
```suggestion
but only submitted to a single queue. The device associated with a queue is not
```

graphs/index.md
```diff
@@ -0,0 +1,706 @@
...
+
+// State of a queue, returned by info::queue::state
+enum class queue_state {
+  executing,
+  recording
+};
+
+// State of a graph
+enum class graph_state {
+  recordable,
```
reble: Naming this state `modifiable` or `editable` might be more generic, and better support future extensions (such as an Explicit API). Or would you consider adding an extra state?

graphs/index.md
```diff
@@ -0,0 +1,706 @@
...
+
+An instance of a `command_graph` object can be in one of two states:
+
+* Recording - Graph is under construction and new nodes may be added to it.
+* Executable - Graph topology is fixed after finalization and graph is ready to
+  be submitted for execution.
+
+A `command_graph` object is constructed in the *recording* state and is made
+*executable* by the user invoking `command_graph::finalize()` to create a
+new executable instance of the graph. An *executable* graph cannot transition
```
reble: Transition could be misleading here. My understanding is, you keep the *editable* `command_graph` object alive, and create (multiple) *executable* object(s) from it, correct? Can `finalize()` method be `const`?

graphs/index.md
EwanC: We weren't considering adding extra state, so calling `recordable` more generic would indeed to better introducing an explicit API. My worry with `modifiable` or `editable` is that the user can perform a kind of `modify` or `edit`  of the graph when we update it, either through whole-graph update or individual node update in an Explicit API.

Would probably tend towards `constructable`, `buildable`, `composable`, or something along those lines. Will have a think.

graphs/index.md
EwanC: > My understanding is, you keep the editable command_graph object alive, and create (multiple) executable object(s) from it, correct?

Yup, you understand correctly. *transition* does indeed sound like there is state change happening in-place, rather than creating something new. Will revisit this wording.

> `Can finalize()` method be const?

Not as we've currently implemented it, but conceptually I think this should be possible, and would make it clearer to the user about the point above of not modifying state in-place. Will take a look at refactoring that section of code to make this happen.

graphs/index.md
EwanC: Decided to make this `modifiable` for the moment to reduce the diff with the oneAPI extension, once they are merged we can always revisit this.

graphs/index.md
EwanC: We tweaked our implementation and indeed this can be `const`, updated to specify that

